PR #2, opened 2017-06-16
releasePublishArtifactsAction should publish Docker image

rossabaker:
It's defaults to `publishSigned`, which is still publishing to Maven.  We want to just execute the `publish` step (same as sbt-release's default) in the http module in order to publish to the Docker registry.


kaiserpelagic:
LGTM

PR #4, opened 2017-06-19
etymology?

tpolecat:
I assume you didn't mean *entomology* 😉


rossabaker:
Right.  The entomology belongs in the issue tracker. :)

PR #5, opened 2017-06-19
Script to sync current branch with upstream master

rossabaker:
Generates a merge commit that makes the current branch identical to the upstream
master without rewriting the history of the current branch.


timperrett:
@rossabaker can you amend the last commit message to [ci skip] so we don't cut a release for this. Thanks

PR #18, opened 2017-07-10
Logging sidecar should be optional 

timperrett:


core/src/main/scala/scheduler/NomadHttp.scala
```diff
@@ -444,8 +454,8 @@ object NomadJson {
             "Count"         := env.desiredInstances.getOrElse(1),
             "RestartPolicy" := env.retries.map(restartJson).getOrElse(Json.jNull),
             "EphemeralDisk" := ephemeralDiskJson(false,false,101),
-            "Tasks"         := List(leaderTaskJson(name,unitName,i,env,BridgeMode,ports,nomad,ns,plan.name, tags), 
-                                loggingSidecarJson(nomad, env.bindings, name, ns))
+            "Tasks"         := (List(leaderTaskJson(name,unitName,i,env,BridgeMode,ports,nomad,ns,plan.name, tags)) ++
```
timperrett: @kaiserpelagic so not exactly - in this case, the `Option` when appended to a list using `++` will actually convert `Option` to `List`. For example:

```
scala> List(1,2,3) ++ Option(4)
res0: List[Int] = List(1, 2, 3, 4)
```

PR #19, opened 2017-07-11
support both v1 and v2 lb protocols

kaiserpelagic:


core/src/main/scala/loadbalancers/Inbound.scala
```diff
@@ -0,0 +1,25 @@
...
+//:   Unless required by applicable law or agreed to in writing, software
+//:   distributed under the License is distributed on an "AS IS" BASIS,
+//:   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//:   See the License for the specific language governing permissions and
+//:   limitations under the License.
+//:
+//: ----------------------------------------------------------------------------
+package nelson
+package loadbalancers
+
```
timperrett: nit: random white space

PR #23, opened 2017-07-26
h2 connection pools; suppress lbs deploy on release

kaiserpelagic:


core/src/main/scala/storage/hikari.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import doobie.contrib.hikari.hikaritransactor._
+import scalaz.concurrent.Task
+
+object Hikari {
+
+  def build(db: DatabaseConfig): HikariTransactor[Task] = {
+    val trans = for {
+      xa <- HikariTransactor[Task](db.driver, db.connection, db.username.getOrElse(""), db.password.getOrElse(""))
+       _ <- xa.configure(hx => Task.delay(db.maxConnections.foreach(max => hx.setMaximumPoolSize(max))))
```
rossabaker: What happens if this is not set?  Does hikari provide a reasonable default?

core/src/test/scala/TestStorage.scala
```diff
@@ -16,16 +16,16 @@
...
   def dbConfig(testname: String): DatabaseConfig =
-    DatabaseConfig("org.h2.Driver", s"jdbc:h2:file:db/nelson.test.$testname.db;DATABASE_TO_UPPER=FALSE;", None, None)
+    DatabaseConfig("org.h2.Driver", s"jdbc:h2:file:db/nelson.test.$testname.db;DATABASE_TO_UPPER=FALSE;", None, None, None)
 
   def storage(testname: String) = {
-    val s = new nelson.storage.H2Storage(dbConfig(testname))
-    nelson.storage.run(s, StoreOp.migrate).run
-    s
+    val cfg = dbConfig(testname)
+    Migrate.migrate(cfg).run
```
rossabaker: I'm still not fond of this run at not-the-end-of-the-world, but it's no worse.

core/src/main/scala/storage/hikari.scala
kaiserpelagic: That's a good question, let me investigate.

kaiserpelagic: default is 10, which is a very sane default.

core/src/test/scala/TestStorage.scala
kaiserpelagic: agreed, was just trying to keep parity with what was already there

core/src/main/scala/storage/hikari.scala
rossabaker: I'm good with that.

timperrett: 👍

PR #24, opened 2017-07-28
expiration policies should work with matrix deployments

kaiserpelagic:


core/src/test/scala/ExpirationPolicySpec.scala
```diff
@@ -176,7 +176,7 @@ class ExpirationPolicySpec extends NelsonSuite with BeforeAndAfterEach with Prop
     RetainLatest.policy(dep411, g)(ext) should equal (None)
   }
 
-  it should "retain the latest by deploy time when version is the same" in {
```
timperrett: What happens here if I deploy a service multiple times using `nelson stack redeploy` ?

kaiserpelagic: then you will have multiple running stacks of the same version

PR #25, opened 2017-07-31
Githubs with minimal repos will hit an NPE

timperrett:


timperrett:
I believe @techlover10 is trying this out locally on his setup, so if he would be so kind: @techlover10 can you report your outcome and let me know if this fixes it for you? Cheers

techlover10:
So interestingly enough, this has the same effect as my extremely hacky fix had earlier - it no longer errors, but it does not find any repos despite the fact that the CURL command does return a JSON list of repos.

PR #29, opened 2017-08-02
fix routing bug

kaiserpelagic:


core/src/main/scala/routing/discovery.scala
```diff
@@ -77,11 +77,11 @@ object Discovery {
...
   }
 
-  def discoveryTables[F[_]: Foldable](graphs: F[(Namespace, RoutingGraph)]): (RoutingNode,NamespaceName) ==>> DiscoveryTables = {
-    graphs.foldLeft[(RoutingNode,NamespaceName) ==>> DiscoveryTables](==>>.empty){(smap,g) =>
+  def discoveryTables[F[_]: Foldable](graphs: F[(Namespace, RoutingGraph)]): (StackName,NamespaceName) ==>> DiscoveryTables = {
+    graphs.foldLeft[(StackName,NamespaceName) ==>> DiscoveryTables](==>>.empty){(smap,g) =>
       val (ns, rg) = g
       rg.nodes.filter(_.nsid == ns.id).foldLeft(smap){(s,rn) =>
-        s.updateAppend((rn, ns.name), discoveryTable(rn, rg))
+        s.insert((rn.stackName, ns.name), discoveryTable(rn, rg))
```
rossabaker: What made this an `insert` instead of an `updateAppend`?

kaiserpelagic: I'm not sure why this was ever `updateAppend` in the first place as a stack should only be in the routing graph once.

kaiserpelagic:
The fix here is to key off of the `StackName` instead of the `RoutingNode` as it uses the underlying `Deployment`'s, `Order` instance for `Map` (`==>>`) operations and that is a bit dubious.

PR #34, opened 2017-08-30
Fix a subtle bug caused by any2string

timperrett:


timperrett:
@rossabaker yeah just some cleanup

PR #37, opened 2017-09-01
fix stackoverflow error when constructing graph 

kaiserpelagic:
`Free.flatMap` is not stacksafe in the version of scalaz that we use. When constructing a routing graph the only time we use a `StoreOpF` is to get the namespace for a loadbalancer. All other calls to StoreOpF are outside the GraphBuild RWST. So the fix for this was to include the namespace on the loadbalancer datatype (just like we do for regular deployments) and use a stacksafe monad.


kaiserpelagic:
Closing this because scalaz 7.1 has several stack safety issues. The real fix here is to upgrade.

PR #49, opened 2017-12-05
Kubernetes integration for Nelson

adelbertc:
* Add Kubernetes interpreter for SchedulerOp
* Add new Canopus workflow for Kubernetes deployments
* Add new 'scheduler' field in config to choose between Nomad or Kubernetes
* Add support for SSLContext in HTTP4S Client creation
* Add namespace to Summary ctor since Kubernetes needs that

Also:
* Change Alpine APK endpoint as original one didn't seem to work anymore
* Remove Allocations, was only used in reconcilliation which was removed in #48

### DIFF
--- a/core/src/main/scala/Config.scala
+++ b/core/src/main/scala/Config.scala
@@ -16,7 +16,11 @@
 //: ----------------------------------------------------------------------------
 package nelson
 
+import java.io.FileInputStream
+import javax.net.ssl.{SSLContext, X509TrustManager}
 import java.nio.file.{Path, Paths}
+import java.security.SecureRandom
+import java.security.cert.{CertificateFactory, X509Certificate}
 import java.util.concurrent.{ExecutorService, Executors, ScheduledExecutorService, ThreadFactory}
 
 import journal.Logger
@@ -32,6 +36,7 @@ import notifications.{SlackHttp,SlackOp,EmailOp,EmailServer}
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
+import scala.util.control.NonFatal
 import scalaz.Scalaz._
 import scalaz.concurrent.Strategy
 import scalaz.~>
@@ -530,17 +535,69 @@ object Config {
         )((a,b,c,d,e,g) => {
           val splunk = readSplunk
           val loggingSidecar = readLoggingImage
-          val uri = org.http4s.Uri.fromString(a).toOption.yolo(s"nomad.endpoint -- $a -- is an invalid Uri")
+          val uri = Uri.fromString(a).toOption.yolo(s"nomad.endpoint -- $a -- is an invalid Uri")
           Infrastructure.Nomad(uri,b,c,d,e,loggingSidecar,g,splunk)
         })
     }
 
-    /*
-     * Datacenters currently only support one scheduler
-     */
-    def readScheduler(kfg: KConfig, proxy: Option[Infrastructure.ProxyCredentials]): Option[SchedulerOp ~> Task] =
-      readNomadInfrastructure(kfg.subconfig("nomad"))
+    def readKubernetesInfrastructure(kfg: KConfig): Option[Infrastructure.Kubernetes] = {
+      (kfg.lookup[String]("endpoint") |@| kfg.lookup[Duration]("timeout")) { (endpoint, timeout) =>
+        val uri = Uri.fromString(endpoint).toOption.yolo(s"kubernetes.endpoint -- $endpoint -- is an invalid Uri")
+        Infrastructure.Kubernetes(uri, timeout)
+      }
+    }
+
+    def readNomadScheduler(kfg: KConfig): Option[SchedulerOp ~> Task] =
+      readNomadInfrastructure(kfg)
         .map(n => new scheduler.NomadHttp(nomadcfg, n, http4sClient(n.timeout)))
+
+    // Create a X509TrustManager that trusts a whitelist of certificates, similar to `curl --cacert`
+    def cacert(certs: Array[X509Certificate]): X509TrustManager =
+      new X509TrustManager {
+        def getAcceptedIssuers(): Array[X509Certificate] = certs
+        def checkClientTrusted(certs: Array[X509Certificate], authType: String): Unit = ()
+        def checkServerTrusted(certs: Array[X509Certificate], authType: String): Unit = ()
+      }
+
+    def getKubernetesPodCert(): Option[SSLContext] = {
+      // Auto-mounted at this path for pods
+      // https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
+      val certBundle = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+      val is = new FileInputStream(certBundle)
+      val x509Cert: Option[X509Certificate] = try {
+        val cf = CertificateFactory.getInstance("X.509")
+        cf.generateCertificate(is) match {
+          case c: X509Certificate => Some(c)
+          case _                  => None
+        }
+      } catch {
+        case NonFatal(_) => None
+      } finally {
+        is.close()
+      }
+
+      x509Cert.map { cert =>
+        val trustManager = cacert(Array(cert))
+        val sslContext = SSLContext.getInstance("TLS")
+        sslContext.init(null, Array(trustManager), new SecureRandom())
+        sslContext
+      }
+    }
+
+    def readKubernetesClient(kfg: KConfig): Option[KubernetesClient] =
+      (readKubernetesInfrastructure(kfg) |@| getKubernetesPodCert()) {
+        case (kubernetes, sslContext) =>
+          // Auto-mounted at this path for pods
+          // https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
+          val path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+          val serviceAccountToken = scala.io.Source.fromFile(path).getLines.toList.head
+          new KubernetesClient(
+            kubernetes.endpoint,
+            http4sClient(kubernetes.timeout, sslContext = Some(sslContext)),
+            serviceAccountToken
+          )
+      }
 
     @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NoNeedForMonad"))
     def readDatacenter(id: String, kfg: KConfig): Datacenter = {
@@ -583,8 +640,19 @@ object Config {
           InstrumentedVaultClient(endpoint, rawClient)
         }).yolo("We really really need vault.  Seriously vault must be configured")
 
-      val sched = readScheduler(kfg.subconfig("infrastructure.scheduler"), proxyCreds)
-        .yolo("At least one scheduler must be defined per datacenter")
+      val schedConfig = kfg.subconfig("infrastructure.scheduler")
+
+      val (sched, healthChecker) = (schedConfig.lookup[String]("scheduler") match {
+        case Some("nomad") =>
+          val s = readNomadScheduler(schedConfig.subconfig("nomad"))
+          val h = health.Http4sConsulHealthClient(consul)
+          s.map((_, h))
+        case Some("kubernetes") =>
+          readKubernetesClient(schedConfig.subconfig("kubernetes")).map { client =>
+            (new scheduler.KubernetesHttp(client), health.KubernetesHealthClient(client))
+          }
+        case _ => None
+      }).yolo("At least one scheduler must be defined per datacenter")
 
       val interpreters = Infrastructure.Interpreters(
         scheduler = sched,
@@ -594,7 +662,7 @@ object Config {
         logger = logger,
         docker = dockerClient,
         control = WorkflowControlOp.trans,
-        health = health.Http4sConsulHealthClient(consul)
+        health = healthChecker
       )
 
       val trafficShift = readTrafficShift(kfg.subconfig("traffic-shift"))
@@ -775,9 +843,11 @@ object Config {
       pkiPath = cfg.lookup[String]("pki-path")
     )
 
-  private def http4sClient(timeout: Duration, maxTotalConnections: Int = 10): Client = {
+  private def http4sClient(timeout: Duration, maxTotalConnections: Int = 10, sslContext: Option[SSLContext] = None): Client = {
     val config = BlazeClientConfig.defaultConfig.copy(
-      requestTimeout = timeout)
+      requestTimeout = timeout,
+      sslContext = sslContext
+    )
     PooledHttp1Client(maxTotalConnections = maxTotalConnections, config = config)
   }
 }
--- a/core/src/main/scala/Datacenter.scala
+++ b/core/src/main/scala/Datacenter.scala
@@ -16,28 +16,29 @@
 //: ----------------------------------------------------------------------------
 package nelson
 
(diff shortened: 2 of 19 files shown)


timperrett:
@rossabaker can you comment on the stability of #35 and it's mergability status?

core/src/main/scala/Config.scala
```diff
@@ -530,18 +535,73 @@ object Config {
...
+      }
+    }
+
+    def readK8sScheduler(kfg: KConfig): Option[SchedulerOp ~> Task] =
+      (readK8sInfrastructure(kfg) |@| getK8sPodCert()) {
+        case (k8s, sslContext) =>
+          // Auto-mounted at this path for pods
+          // https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
+          val path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+          val serviceAccountToken = scala.io.Source.fromFile(path).getLines.toList.head
```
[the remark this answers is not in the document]
adelbertc: Currently yeah, which implies either single datacenter/region or awkward things like having a blessed master region that has line of sight to every other region. I should get it working w/ plain Docker like there is for Nomad, but this was done mostly as part of my "let's just get it working" effort

core/src/main/scala/Config.scala
```diff
@@ -530,18 +535,73 @@ object Config {
...
+          new scheduler.K8sHttp(k8s, http4sClient(k8s.timeout, sslContext = Some(sslContext)), serviceAccountToken)
+      }
+
+    def readScheduler(kfg: KConfig, proxy: Option[Infrastructure.ProxyCredentials]): Option[SchedulerOp ~> Task] =
+      kfg.lookup[String]("scheduler") match {
+        case Some("nomad") => readNomadScheduler(kfg.subconfig("nomad"))
+        case Some("k8s")   => readK8sScheduler(kfg.subconfig("k8s"))
```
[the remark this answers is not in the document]
adelbertc: Sure :-) Did you want this change just in the config or for the classes as well, e.g. `s/K8sHttp/KubernetesHttp`

core/src/main/scala/Datacenter.scala
```diff
@@ -16,27 +16,28 @@
...
-import scalaz.syntax.monoid._
-import scalaz.std.set._
-import scalaz.syntax.std.option._
-import scalaz.syntax.foldable._
-import scalaz.concurrent.Task
 import java.net.URI
 import java.time.Instant
 import scala.concurrent.duration.FiniteDuration
 
+import com.amazonaws.regions.Region
```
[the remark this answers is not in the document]
adelbertc: This happened since I was moving around/sorting the imports, my PR specifically does not use it but it's being used somewhere else in the file.

core/src/main/scala/scheduler/K8sHttp.scala
```diff
@@ -0,0 +1,156 @@
...
+      Task.delay {
+        log.debug(s"K8sHttp Summary(${dc}, ${stackName})")
+        None
+      }
+    case RunningUnits(dc, prefix) =>
+      Task.delay {
+        log.debug(s"K8sHttp RunningUnits(${dc}, ${prefix})")
+        Set.empty
+      }
+    case Allocations(dc, prefix) =>
```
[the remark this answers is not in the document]
adelbertc: Good point, I'll have a change to completely remove it

core/src/main/scala/scheduler/K8sHttp.scala
```diff
@@ -0,0 +1,156 @@
...
+    name:     String,
+    // unitName: UnitName,          // ignored
+    image:    Image,
+    // dc:       Datacenter,        // ignored
+    // ns:       NamespaceName,     // ignored
+    plan:     Plan,
+    // schedule: Option[Schedule],  // ignored
+    ports:    Option[Ports]
+  ): Json = {
+    // TODO: Cron/Job not supported atm, only Service
```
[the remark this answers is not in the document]
adelbertc: Yes sir, I just haven't done it yet, I just wanted to get something working so I focused on services. But absolutely we can supports jobs!

core/src/main/scala/scheduler/K8sHttp.scala
```diff
@@ -0,0 +1,156 @@
...
+              "nelson" := "true"
+            )
+          ),
+          "spec" := Json(
+            "containers" := List(
+              Json(
+                "name"  := name,
+                "image" := image.toString,
+                "ports" := List(
+                  Json("containerPort" := ports.map(_.default.port).getOrElse(8081))
```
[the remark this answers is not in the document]
adelbertc: Fixed in incoming PR :-)

coltfred:
I'm super pumped you did this @adelbertc. I'll need to look at it in more detail later.

core/src/main/scala/scheduler/K8sHttp.scala
```diff
@@ -0,0 +1,156 @@
...
+        "name"   := name,
+        "labels" := Json(
+          "app"    := name,
+          "nelson" := "true"
+        )
+      ),
+      "spec" := Json(
+        "replicas" := plan.environment.desiredInstances.getOrElse(1),
+        "selector" := Json(
+          "matchLabels" := Json("app" := name)
```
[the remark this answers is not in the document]
adelbertc: Kubernetes has a notion of [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) for each pod (analogous to a unit of deployment, could be a service, a service + sidecars), this `matchLabels` determines which pods are controlled by this deployment. This becomes important since the pods deployed through a Kubernetes Deployment aren't special, there could be existing pods that match a label in which case the Deployment will consider those as part of it's control.

For example if there is an existing `foo` with label `name=foo`, and you create a Deployment for `foo` with `"matchLabels": { "name": "foo" }` for 2 replicas, it will notice there is already an existing one and only create one additional pod.

tl;dr `matchLabels` should match on labels that uniquely identify a particular deployment, so I think `name` (which is the stack name) makes sense here, do I need anything else?

core/src/main/scala/scheduler/K8sHttp.scala
```diff
@@ -0,0 +1,156 @@
...
+        ),
+        "template" := Json(
+          "metadata" := Json(
+            "labels" := Json(
```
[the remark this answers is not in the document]
adelbertc: For sure, I've added stack name, unit name, and version in an incoming commit.

adelbertc:
@coltfred Yes! Is your company using Kubernetes? I'm planning on writing a "Setup Nelson for K8s" guide after we finish up this initial PR. Also feel free to ask me (or anyone else) questions in the Gitter channel.

daddykotex:
👍 I'm impressed by how little code it took. Really looking forward to it. Thanks for that

core/src/main/scala/Config.scala
```diff
@@ -530,18 +535,73 @@ object Config {
...
+      } catch {
+        case NonFatal(_) => None
+      } finally {
+        is.close()
+      }
+
+      x509Cert.map { cert =>
+        val trustManager = cacert(Array(cert))
+        val sslContext = SSLContext.getInstance("TLS")
+        sslContext.init(null, Array(trustManager), new SecureRandom())
```
timperrett: Doesn't this make a new secure random for every request? It's not clear if this code block is called frequently or not.

core/src/main/scala/Config.scala
```diff
@@ -530,18 +535,73 @@ object Config {
...
+      (readKubernetesInfrastructure(kfg) |@| getKubernetesPodCert()) {
+        case (kubernetes, sslContext) =>
+          // Auto-mounted at this path for pods
+          // https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
+          val path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+          val serviceAccountToken = scala.io.Source.fromFile(path).getLines.toList.head
+          new scheduler.KubernetesHttp(kubernetes, http4sClient(kubernetes.timeout, sslContext = Some(sslContext)), serviceAccountToken)
+      }
+
+    def readScheduler(kfg: KConfig, proxy: Option[Infrastructure.ProxyCredentials]): Option[SchedulerOp ~> Task] =
```
timperrett: Can we just double check there's no documentation that needs updating pertaining to this?

core/src/main/scala/scheduler/DeploymentSummary.scala
```diff
@@ -17,10 +17,28 @@
 package nelson
 package scheduler
 
+import scalaz.Monoid
+import scalaz.Scalaz._
```
timperrett: Can we use the a la carte imports within the scope of a module? I dont like importing these globally in a file

core/src/main/scala/scheduler/KubernetesHttp.scala
```diff
@@ -0,0 +1,407 @@
...
+  def apply[A](fa: SchedulerOp[A]): Task[A] = fa match {
+    case Delete(dc, deployment) =>
+      delete(dc, deployment)
+    case Launch(image, dc, ns, unit, plan, hash) =>
+      launch(image, dc, ns, Versioned.unwrap(unit), unit.version, plan, hash)
+    case Summary(dc, ns, stackName) =>
+      summary(dc, ns, stackName)
+
+    // TODO: Legacy, no longer used
+    case RunningUnits(dc, prefix) =>
```
timperrett: This is actually used in `nelson stack runtime` I believe :-)

core/src/main/scala/scheduler/KubernetesHttp.scala
```diff
@@ -0,0 +1,407 @@
...
+          "version"     := stackName.version.toString,
+          "nelson"      := "true"
+        )
+      )
+    )), jobSpecJson(stackName, image, plan, ports)))
+
+  private def jobSpecJson(
+    stackName: StackName,
+    image:     Image,
+    // dc:       Datacenter,        // ignored
```
timperrett: Zombie code

core/src/main/scala/Config.scala
adelbertc: Appears so yeah, but it should only be called once on startup I think since (currently) it's only being used to create the K8s scheduler interpreter. I could move it out but I'm not sure what the security or mutability ramifications of that are.

core/src/main/scala/scheduler/KubernetesHttp.scala
adelbertc: Hm I tried tracing it in the codebase and also looking at https://verizon.github.io/nelson/reference.html#cli , I think that only uses `Summary` ? It seems `RunningUnits` is mostly used in `NomadHttp` so it can kill child processes for jobs and stuff ?

adelbertc: My trace:

* [`/runtime` endpoint](https://github.com/Verizon/nelson/blob/master/http/src/main/scala/plans/Datacenters.scala#L234) calls `Nelson.getRuntimeSummary`
* [`Nelson.getRuntimeSummary`](https://github.com/Verizon/nelson/blob/master/core/src/main/scala/Nelson.scala#L592) calls `SchedulerOp.summary`

core/src/main/scala/Config.scala
adelbertc: Just checked again, I did a grep through the repo for `nomad {` and made sure there was a `scheduler = "nomad"` in those locations:

```grep
$ rg "nomad \{" --context 1
etc/development/http/http.dev.cfg
93-          scheduler = "nomad"
94:          nomad {
95-           endpoint = "http://localhost:4646"
--
117-      }
118:      # nomad {
119-      #   endpoint = "$(NOMAD_ADDR)"

etc/development/docker/nelson.cfg
47-          scheduler = "nomad"
48:          nomad {
49-            endpoint = ""

docs/src/hugo/content/index.md
689-          scheduler = "nomad"
690:          nomad {
691-            # Where can Nelson access the Nomad leader cluster

core/src/test/resources/nelson/datacenters-missing-consul.cfg
33-          scheduler = "nomad"
34:          nomad {
35-            endpoint  = "https://nomad.service"
--
78-          scheduler = "nomad"
79:          nomad {
80-            endpoint  = "https://nomad.service"

core/src/test/resources/nelson/datacenters-missing-domain.cfg
32-          scheduler = "nomad"
33:          nomad {
34-            endpoint  = "https://nomad.service"

core/src/test/resources/nelson/datacenters.cfg
32-          scheduler = "nomad"
33:          nomad {
34-            endpoint  = "https://nomad.service"
--
87-          scheduler = "nomad"
88:          nomad {
89-            endpoint  = "https://nomad.service"

core/src/main/resources/nelson/defaults.cfg
135-
136:  nomad {
137-    # These service tags will be automatically added (via set union)
--
236-  #         scheduler = "nomad"
237:  #         nomad {
238-  #           endpoint = "http://nomad.service.texas.your.company.com:1234/"
```

README.markdown
```diff
@@ -0,0 +1,4 @@
+# Nelson has moved!
+
+Please find Nelson at its new home: [getnelson/nelson](https://getnelson.github.io/nelson/)
```
timperrett: @adelbertc is this a bad merge? In this repo we don't want to change the repo from what it was IMHO

timperrett:
@kaiserpelagic @rossabaker @adelbertc so we got the healthcheck stuff on the other PR, but what else will it take to get this over the line? Cheers

README.markdown
adelbertc: Yes definitely a bad merge, will fix...

adelbertc:
So my goal for this PR is to get Nelson running on a single K8s cluster.. with that in mind I think these are the missing pieces:

* Health check algebra implementation
* Service discovery
* Load balancer support

Health check algebra is just a matter of getting the other PR merged and then implementing it, should be straightforward.

For service discovery it seems at least short term we have 2 options:
* Write a key-value store algebra (since currently things are tied to Consul) and use it for Lighthouse - Kubernetes will probably implement this with etcd
* Punt to the routing layer, e.g. teach Envoy how to fetch/read/interpret Lighthouse, if possible

For load balancing we have a couple options:

### Single ingress
Seems to be the predominant pattern in Kubernetes, or rather nobody I've asked has refuted me on that claim yet. This means all requests go to the same IP address, share the same load balancer (including # of nodes handling requests, etc.) and have two options for routing:

* Hostname - every service gets its own hostname, all of which point to the single ingress IP address. Routing rules are defined on a per-hostname basis

```yaml
spec:
  rules:
  - host: foo.bar.com
    http:
      paths:
      - backend:
          serviceName: s1
          servicePort: 80
  - host: bar.foo.com
    http:
      paths:
      - backend:
          serviceName: s2
          servicePort: 80
```

* Routes - all services share a hostname, and are routed via the path, e.g. `<load balancer>/foo/bar`

```yaml
spec:
  rules:
  - host: foo.bar.com
    http:
      paths:
      - path: /foo
        backend:
          serviceName: s1
          servicePort: 80
      - path: /bar
        backend:
          serviceName: s2
          servicePort: 80
```

### Multiple Ingress

Seems to be not that common in Kubernetes, or at least doesn't seem to be a direction the Kubernetes folks are heading. This would however match more closely with Nelson's current design of exposing a load balancer per service.


I don't have many thoughts re: single vs. multiple ingress ATM, I do think many users currently do not operate at a scale/domain where having a single ingress point matters much? At least it doesn't for us, and we can certainly just implement both suggestions I made above and somehow make loadbalancer implementation pluggable. I do agree the multiple loadbalancer thing is ultimately more scalable though.

PR #50, opened 2017-12-07
Remove redundant algebra members

timperrett:


timperrett:
@kaiserpelagic @rossabaker can you check me on this? I think these were only needed for reconciliation

kaiserpelagic:
LGTM

PR #53, opened 2017-12-08
Rename Datacenter to Domain

timperrett:
Here's the work for #52 - im not totally set on the name, so i'd love to get some opinions on this.


rossabaker:
Is there a corresponding nelson-cli change?

Migration path might be eased by preserving the old HTTP endpoints and complicating the config reader, though it's cruft.

This leaves a lot of variable names `dc` and `dcs` that make sense only with historical context, but the types are good.  It's just names.

timperrett:
@rossabaker good thought about the http endpoints, i can add them back in. Will also put up a similar change to the Nelson CLI :-)

resya173:
Hi

PR #55, opened 2017-12-19
health check algebra

kaiserpelagic:
I kept this pretty simple for now by only adding the health check algebra and not changing any of the consul related stuff. That's going to be a bigger change. Also, there's only a single interpreter back by consul but I imagine we'll also want one for k8s.


adelbertc:
LGTM - if you guys are OK with this I can rebase and work in the K8s stuff after.

PR #56, opened 2017-12-20
Make health check algebra even simpler

adelbertc:
@kaiserpelagic


kaiserpelagic:
LGTM

core/src/main/scala/Json.scala
```diff
@@ -602,7 +599,7 @@ object Json {
   implicit lazy val RuntimeSummaryEncoder: EncodeJson[Nelson.RuntimeSummary] =
     EncodeJson((rs: Nelson.RuntimeSummary) =>
       ("scheduler" := rs.deployment) ->:
-      ("consul_health" := rs.health) ->:
+      ("healthy" := rs.healthy) ->:
```
adelbertc: This would be a breaking API change - if we are OK with this we would need to update the Nelson CLI: https://github.com/getnelson/nelson-cli/blob/a0ac4436fd853db60f74d565db446459a31ad2cb/src/github.com/getnelson/nelson/stack.go#L401

adelbertc:
Closing in favor of https://github.com/getnelson/nelson/pull/57

PR #57, opened 2017-12-21
Add more information to health check algebra

adelbertc:
Also change runtime inspet to use health check algebra instead of
hitting Consul directly

cc @kaiserpelagic


core/src/main/scala/health/Http4sConsulHealthClient.scala
```diff
@@ -40,10 +40,11 @@ final case class Http4sConsulHealthClient(client: ConsulOp ~> Task) extends (Hea
     case _          => Unknown
   }
 
-  implicit val healthCheckDecoder: DecodeJson[HealthCheck] =
-    DecodeJson[HealthCheck] { c =>
-      c.as[String].flatMap { s =>
```
kaiserpelagic: Oh nice, this was a bug. It should have been `(c --\ "Status").as[String]`

adelbertc:
@timperrett thoughts? will need this for k8s-y stuff :-)

core/src/main/scala/Json.scala
```diff
@@ -583,21 +583,14 @@ object Json {
   implicit val HealthCheckEncoder: EncodeJson[health.HealthCheck] =
     EncodeJson[health.HealthCheck] { hs => jString(health.HealthCheck.toString(hs)) }
 
-  implicit lazy val HealthStatusEncoder: EncodeJson[ConsulHealthStatus] =
-    EncodeJson((h: ConsulHealthStatus) =>
-      ("name"    := h.name) ->:
+  implicit lazy val HealthStatusEncoder: EncodeJson[HealthStatus] =
+    EncodeJson((h: HealthStatus) =>
+      ("name"    := h.details.getOrElse("")) ->:
```
timperrett: I'm not sure `""` is the best default. I would either prefer "unspecified" or make this field optional.

timperrett:
@adelbertc other than that one change i commented on and would request we change, this looks good!

adelbertc:
We good to go on this?

PR #58, opened 2017-12-25
yaml is cool too

adelbertc:
Corresponding slipway change: https://github.com/getnelson/slipway/pull/2

### DIFF
--- a/core/src/main/scala/Manifest.scala
+++ b/core/src/main/scala/Manifest.scala
@@ -387,7 +387,9 @@ object Manifest {
    * place i could find to put it.
    */
   private def parseDeployable(release: Github.Release, name: String): Task[Deployable] = {
-    release.findAssetContent(s"${name}.deployable.yml").flatMap { a =>
+    release.findAssetContent(s"${name}.deployable.yml").handleWith {
+      case ProblematicDeployable(_, _) => release.findAssetContent(s"${name}.deployable.yaml")
+    }.flatMap { a =>
       yaml.DeployableParser.parse(a).fold(e => Task.fail(MultipleErrors(e)), Task.now)
     }
   }
--- a/core/src/main/scala/Exceptions.scala
+++ b/core/src/main/scala/Exceptions.scala
@@ -44,7 +44,7 @@ final case class ProblematicRepoManifest(slug: Slug)
   extends NelsonError(s"repo '${slug.toString}' was expected to have a valid .nelson.yml")
 
 final case class ProblematicDeployable(str: String, url: String)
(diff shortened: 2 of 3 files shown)


core/src/main/scala/Manifest.scala
```diff
@@ -387,7 +387,9 @@ object Manifest {
    * place i could find to put it.
    */
   private def parseDeployable(release: Github.Release, name: String): Task[Deployable] = {
-    release.findAssetContent(s"${name}.deployable.yml").flatMap { a =>
+    release.findAssetContent(s"${name}.deployable.yml").handleWith {
+      case ProblematicDeployable(_, _) => release.findAssetContent(s"${name}.deployable.yaml")
+    }.flatMap { a =>
```
rossabaker: This is going to make the error message reference `.yaml` instead of `.yml`, which I guess is not a big deal since both are now supported.  I thought about suggesting `findAssetContent` should take a predicate, but that complicates the error message generation.

I think I'm fine with this.  Just thinking out loud.

timperrett: Could just do the lazy:

```
case ProblematicDeployable(_, _) => release.findAssetContent(s"${name}.deployable.yaml or ${name}.deployable.yml")
```

adelbertc: Aha that's an interesting hack, would cause a third useless traversal though the list though. Perhaps it's better to just leave it and decide if we care about `.yml` or `.yaml` more? :P

timperrett: @adelbertc how would it cause an extra traversal? I'm just suggesting altering the message to make it clear for users :-)

adelbertc: The string argument is used [both to find the actual content and for the error message](https://github.com/getnelson/nelson/blob/master/core/src/main/scala/Github.scala#L93), so you can't do what you suggested without either breaking the actual find-the-asset bit
or doing a 3rd dummy traversal for the benefit of getting a better error message

timperrett: Ahhhh sorry i'm being slow. Yeah not ideal I guess - would love something to indicate to the users as this is one of the most common messages first-time users will see. 

Perhaps as an alternitive we could simply augment the underlying error message? https://github.com/getnelson/nelson/blob/master/core/src/main/scala/Exceptions.scala#L47 - how about:

```
NelsonError(s"Error when attempting to parse deployable from repository. Please ensure that a valid deployable yaml/yml file is available at the specified URL, or attached to the associated Github release: ${url} -- $str")
```

PR #62, opened 2018-01-18
More Nelson Kubernetes features + fixes

adelbertc:
Features
* Use manifest environment variables, resource requests, and health
  checks

Fixes
* Job backoffLimit defaults to 3 to match Nomad behavior
* Fix health check for pods - Kubernetes has several "status" fields for
  pods, after some experience in production the new behavior implemented
  in this PR is the more useful one
    * Old behavior would consider the pods "Running" even if the Pod
      deployments were failing, e.g. CrashLoopBackOff. I suppose they
      were "running" in the sense that they're doing something, but
      that's not very useful for health checking
* Set restartPolicy on Jobs to Never due to a bug in Kubernetes

### DIFF
--- a/core/src/main/scala/KubernetesClient.scala
+++ b/core/src/main/scala/KubernetesClient.scala
@@ -15,7 +15,7 @@ import scalaz.concurrent.Task
 import scalaz.Scalaz._
 
 import nelson.Datacenter.StackName
-import nelson.Manifest.{EnvironmentVariable, Plan, Ports, Port}
+import nelson.Manifest.{EnvironmentVariable, HealthCheck => HealthProbe, Plan, Ports, Port}
 import nelson.docker.Docker.Image
 import nelson.health._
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/KubernetesClient.scala
```diff
@@ -173,9 +173,16 @@ object KubernetesJson {
...
-                "name"  := stackName.toString,
-                "image" := image.toString,
-                "ports" := containerPortsJson(ports.toList.flatMap(_.nel.list))
+                "name"      := stackName.toString,
+                "image"     := image.toString,
+                "env"       := plan.environment.bindings,
+                "ports"     := containerPortsJson(ports.toList.flatMap(_.nel.list)),
+                "resources" := argonaut.Json(
+                  "requests" := argonaut.Json(
+                    "cpu"    := plan.environment.cpu.getOrElse(0.5),
```
adelbertc: Oops need to do this for jobs as well..

adelbertc: @timperrett I do wonder, Kubernetes has both a notion of "requests" (lower bound) and "limits" (upper bound). It seems like Nomad's resource stanza is a lower bound with no upper bound? Or is it both. Not having an upper bound seems unsafe.

timperrett: @adelbertc its the other way around - nomad only has upper bounds. Lower bounds don't make sense because then you can't stochastically oversubscribe. So provided you have upper bounds, you should be set.

timperrett:
@adelbertc just that one change from lower bound to upper bound and you should be set

adelbertc:
@timperrett Done! Stuff I changed:

* Changed `requests` to `limits` which is the K8s equivalent of resource upper-bound: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu
* Moved health checks to services, not jobs, oops
* Added resource limits to jobs as well, not just services

PR #64, opened 2018-02-19
s/scalaz.concurrent.Task/cats.effect.IO, s/scalaz.stream.Process/fs2.Stream

adelbertc:


core/src/main/scala/CatsHelpers.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  implicit class NelsonEnrichedEither[A, B](val either: Either[A, B]) extends AnyVal {
+    def toDisjunction: scalaz.\/[A, B] = either match {
+      case Left(a)  => scalaz.-\/(a)
+      case Right(b) => scalaz.\/-(b)
+    }
+  }
+
+  // Adapted from https://github.com/Verizon/delorean/blob/master/core/src/main/scala/delorean/package.scala
+  implicit class NelsonEnrichedFuture[A](future: => Future[A]) {
```
rossabaker: `IO` has an `unsafeFromFuture`.

core/src/main/scala/CatsHelpers.scala
```diff
@@ -0,0 +1,87 @@
...
+    def or(other: IO[A]): IO[A] = io.attempt.flatMap {
+      case Right(a) => IO.pure(a)
+      case Left(e)  => other
+    }
+
+    /** Fail with error if the result of the IO does not satsify the predicate
+      *
+      * Taken from https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+      */
+    def ensure(failure: => Throwable)(f: A => Boolean): IO[A] =
```
rossabaker: I wonder whether this makes sense on `ApplicativeError`.

core/src/main/scala/CatsHelpers.scala
```diff
@@ -0,0 +1,87 @@
...
+      io.flatMap(a => if (f(a)) IO.pure(a) else IO.raiseError(failure))
+
+    def timed(timeout: Duration)(implicit ec: ExecutionContext): IO[A] = for {
```
rossabaker: Can we call this `unsafeTimed` until we figure out how to kill it entirely or do it safely?  This is a scary thing.

core/src/main/scala/Nelson.scala
```diff
@@ -396,11 +408,10 @@ object Nelson {
    * List all the datacenters Nelson is currently aware of, and return
    * the namespaces associated with that datacenter.
    */
-  def listDatacenters: NelsonK[Map[Datacenter, Set[Namespace]]] = Kleisli { cfg =>
-    Nondeterminism[Task].gatherUnordered(
-      cfg.datacenters.map(d =>
-        storage.run(cfg.storage,
-                    storage.StoreOp.listNamespacesForDatacenter(d.name).map(d -> _)))).map(_.toMap)
+  def listDatacenters(implicit ec: ExecutionContext): NelsonK[Map[Datacenter, Set[Namespace]]] = Kleisli { cfg =>
+    parallelTraverse(cfg.datacenters) { d =>
```
adelbertc: Needs a look

core/src/main/scala/Pipeline.scala
```diff
@@ -46,18 +50,22 @@ object Pipeline {
...
    * effect generating sinks and observations.
    */
-  def task(config: NelsonConfig)(effects: Sink[Task, Action]): Task[Unit] = {
+  def task(config: NelsonConfig)(effects: Sink[IO, Action]): IO[Unit] = {
+    def par[A](ps: Stream[IO, Stream[IO, A]]) = {
+      implicit val ec = config.pools.defaultExecutor
 
-    /* needed for parallel consumption of the pipeline */
-    import scalaz.stream.nondeterminism.njoin
+      val withErrors = ps.join(config.pipeline.concurrencyLimit).attempt
```
adelbertc: Needs a look

core/src/main/scala/alerts/Overhaul.scala
```diff
@@ -16,31 +16,36 @@
...
         stackName.toString, "-p", plan, "-e", ns.root.asString)
       pb.redirectErrorStream(true)
       val p = pb.start()
 
-      Pools.default.serverExecutor {
-        val in = new BufferedWriter(new OutputStreamWriter(p.getOutputStream, UTF_8))
-        in.write(rules)
-        in.flush()
-        in.close()
-      }
```
adelbertc: Needs a look - `serverExecutor: Strategy` but `Strategy#apply[A]` gives `() => A` but it never seems to be actually run here?

core/src/main/scala/package.scala
```diff
@@ -75,22 +81,30 @@ package object nelson {
...
+    def tfold[B](e: NelsonError)(f: A => B): IO[B] =
+      in.fold(fail[B](e))(a => IO(f(a)))
   }
 
-  implicit class BedazzledTask[A](in: Task[A]){
-    def retryExponentially(seed: Duration = 15.seconds, limit: Int = 5): Task[A] = {
-      val periods = List.fill(limit)(seed).zipWithIndex.map {
-        case (d,i) => d * i.toDouble
+  implicit class BedazzledIO[A](in: IO[A]){
+    def retryExponentially(seed: FiniteDuration = 15.seconds, limit: Int = 5)(scheduler: ScheduledExecutorService, ec: ExecutionContext): IO[A] = {
```
adelbertc: Needs a look

core/src/main/scala/monitoring/Stoplight.scala
```diff
@@ -16,29 +16,25 @@
...
 
-  def apply[A](gauge: Gauge)(p: Process[Task, A]): Process[Task, A] =
-    Process.await(Task.delay(gauge.set(Running))) { _ =>
-      p.onHalt {
-        case Cause.End => Process.eval_(Task.delay(gauge.set(Stopped)))
-        case _ => Process.eval_(Task.delay(gauge.set(Failed)))
-      }
-    }
+  def apply[A](gauge: Gauge)(p: Stream[IO, A]): Stream[IO, A] =
+    (Stream.eval_(IO(gauge.set(Running))) ++ p ++ Stream.eval_(IO(gauge.set(Stopped)))).handleErrorWith(_ => Stream.eval_(IO(gauge.set(Failed))))
```
adelbertc: Needs a look

http/src/main/scala/Main.scala
```diff
@@ -34,33 +40,35 @@ object Main {
...
     val file = new File(args.headOption.getOrElse("/opt/application/conf/nelson.cfg"))
 
-    def readConfig =
-      (knobs.loadImmutable(Required(ClassPathResource("nelson/defaults.cfg")) :: Nil) |@|
-        knobs.loadImmutable(Optional(FileResource(file)) :: Nil))((a,b) =>
-        Config.readConfig(a ++ b, Http, Hikari.build _)
-      )
+    def readConfig = for {
+      defaults  <- knobs.loadImmutable[IO](Required(ClassPathResource("nelson/defaults.cfg")) :: Nil)
+      overrides <- knobs.loadImmutable[IO](Optional(FileResource(file)(singleThreadPool)) :: Nil)
```
adelbertc: Needs a look - use of `singleThreadPool`

http/src/main/scala/Main.scala
```diff
@@ -69,11 +77,11 @@ object Main {
       import cleanup.SweeperDefaults._
 
       log.info("booting the background processes nelson needs to operate...")
-      runBackgroundJob("auditor", cfg.auditor.process(cfg.storage))
-      runBackgroundJob("pipeline_processor", Process.eval(Pipeline.task(cfg)(Pipeline.sinks.runAction(cfg))))
+      runBackgroundJob("auditor", Stream.force(cfg.auditor.map(auditor => auditor.process(cfg.storage)(cfg.pools.defaultExecutor))))
```
adelbertc: Needs a look - use of `defaultExecutor`

http/src/main/scala/Main.scala
```diff
@@ -69,11 +77,11 @@ object Main {
...
+      runBackgroundJob("pipeline_processor", Stream.eval(Pipeline.task(cfg)(Pipeline.sinks.runAction(cfg))))
       runBackgroundJob("workflow_logger", cfg.workflowLogger.process)
       runBackgroundJob("routing_cron", routing.cron.consulRefresh(cfg) to Http4sConsul.consulSink)
-      runBackgroundJob("cleanup_pipeline", cleanup.CleanupCron.pipeline(cfg))
+      runBackgroundJob("cleanup_pipeline", cleanup.CleanupCron.pipeline(cfg)(cfg.pools.defaultExecutor))
```
adelbertc: Needs a look - use of `defaultExecutor`

http/src/main/scala/Accepts.scala
```diff
@@ -16,13 +16,14 @@
...
+import cats.effect.IO
 import org.http4s._
 
 /** This should be added to http4s */
 abstract class Accepts(mediaType: MediaType) {
-  def unapply(req: Request): Boolean =
+  def unapply(req: Request[IO]): Boolean =
     req.headers.get(headers.Accept).fold(true)(
-      _.values.list.exists(_.mediaRange.satisfiedBy(mediaType)))
+      _.values.toList.exists(_.mediaRange.satisfiedBy(mediaType)))
```
rossabaker: Doesn't `Foldable` have `exists` now?  Not that it's a showstopper.

http/src/main/scala/plans/UI.scala
```diff
@@ -53,71 +57,71 @@ final case class UI(config: NelsonConfig) extends Default {
...
+      Found(Location(uri("/login")))
   }
 
   /*
    * if the server config has been overriden to provide an fs path
    * then we'll try to serve the file from that location, otherwise,
    * we'll use the built-in nelson-ui that comes from the classpath dependency.
    */
-  private def serveStaticFile(path: String) =
+  private def serveStaticFile(path: String): Response[IO] =
```
rossabaker: This seems like an `IO[Response[IO]]` to me.

adelbertc: Wdym? It type checks? :D

http/src/main/scala/Accepts.scala
adelbertc: `scalaz.NonEmptyList` so its either here or define the instance in `CatsHelpers`. Since eventually it's going to become `cats.data.NonEmptyList` I figure I'd touch the minimal amount of stuff

core/src/main/scala/audit/Auditor.scala
```diff
@@ -17,43 +17,51 @@
...
+    Sink(t => IO(logger.error(t.getMessage))) // possibly truncate
 
-  def write[A](a: A, action: AuditAction, releaseId: Option[Long] = None, login: String = defaultLogin)(implicit au: Auditable[A]): Task[Unit] =
-    queue.enqueueOne(AuditEvent(a, action, releaseId, login))
+  def write[A](a: A, action: AuditAction, releaseId: Option[Long] = None, login: String = defaultLogin)(implicit au: Auditable[A]): IO[Unit] =
+    queue.enqueue1(AuditEvent(a, action, releaseId, login))
 
-  def process(stg: (StoreOp ~> Task)): Process[Task, Unit] =
-    (queue.dequeue.observe(persist(stg))).attempt().stripW.to(logSink)
+  def process(stg: (StoreOp ~> IO))(implicit ec: ExecutionContext): Stream[IO, Unit] =
```
kaiserpelagic: What is the `ExecutionContext` used for here?

http/src/main/scala/plans/UI.scala
rossabaker: I'm saying I think this should return an `IO[Response[IO]]` because it's, like, doing `IO`.  And then a bunch of those other changes you made melt away.

rossabaker:
Was there a force push?  I'm trying to look at the "needs a look," but every time I click through, GitHub says it can't find the commit.  I'm not seeing any more context than a few lines.

adelbertc:
@rossabaker I don't think so? I see it on my end at least.. does it help if you switch to the "Files changed" tab?

http/src/main/scala/plans/UI.scala
adelbertc: Ohh I see, yeah you're right. 👍

core/src/main/scala/audit/Auditor.scala
adelbertc: `Stream#observe` needs an `ExecutionContext` in FS2 due to differences in encoding - I asked this same question in the FS2 Gitter :-) See: https://gitter.im/functional-streams-for-scala/fs2?at=5a877d3d0202dc012e44b0a8

http/src/main/scala/plans/UI.scala
adelbertc: @rossabaker Hmm taking a closer look now I'm less sure. I assume by "doing `IO`" you are referring to `fileFromFilesystem` and `fileFromClasspath`, both of which give back `Response[IO]` - any `IO` being done in there is done through FS2 which gives a `fs2.Stream[IO, _]` and stuffed in a `Response` which gives `Response[IO]` not `IO[Response[IO]]`. Unless this is a bug with how HTTP4S let's you construct `Response`s?

core/src/main/scala/Metrics.scala
```diff
@@ -156,13 +159,17 @@ object Metrics {
...
+    def apply[A](op: IO[A]) : IO[A] = {
+      val io = for {
+        _       <- before
+        start   <- IO(System.nanoTime())
+        a       <- op
+        elapsed = (System.nanoTime - start) / 1.0e9
+        _       <- onComplete.run(elapsed)
+      } yield a
+
+      (io <* onSuccess).onError { case err => onFail(err) }
```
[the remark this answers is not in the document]
rossabaker: Where does `onError` come from?  Why is this not a `handleWith`?

core/src/main/scala/Pipeline.scala
rossabaker: :+1:

core/src/main/scala/monitoring/Stoplight.scala
rossabaker: :+1:

core/src/main/scala/package.scala
rossabaker: I'm not familiar with this `retry` on fs2, but sees legit.

http/src/main/scala/plans/UI.scala
rossabaker: Oh, I see.  I thought you were using http4s' staticcontent helpers, which return an `OptionT[F, Response[F]]`.

So this method of constructing a response is constructing 200 OK responses that might fail.  You might want to do an `IO` to ensure that the file is there and readable and then return the stream, or else return a 404.  But what you've done is also legit.

http/src/main/scala/Main.scala
rossabaker: I think I was thinking it would be good to name the thread after what it's doing, instead of an opaque thread name.

core/src/main/scala/Metrics.scala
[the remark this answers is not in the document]
adelbertc: `onError` is from `ApplicativeError`: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/ApplicativeError.scala#L99-L128

I believe the use here is primarily convenience, we just want to run a handler and then give back the original result. Could do it with `handleErrorWith` with some dancing about. I also wonder if it might be better to just do something like:

```scala
io.attempt.flatMap {
  case Left(e) => onFail(e) *> IO.raiseError(e)
  case Right(a) => onSuccess(a) *> IO.pure(a)
}
```

As far as I know this is what the Scalaz equivalent of `onFinish` was doing: https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Task.scala#L50-L61

I'm also now wondering why it isn't just implemented the way I did above..

fommil:
you could also have migrated to https://github.com/scalaz/ioeffect/ without the laziness risk.

PR #67, opened 2018-03-22
Add multi-DC support for Kubernetes integration

adelbertc:
Also add support for Kubernetes versions 1.6.x and 1.7.x.

### DIFF
--- a/core/src/main/scala/Config.scala
+++ b/core/src/main/scala/Config.scala
@@ -540,12 +540,16 @@ object Config {
         })
     }
 
-    def readKubernetesInfrastructure(kfg: KConfig): Option[Infrastructure.Kubernetes] = {
-      (kfg.lookup[String]("endpoint") |@| kfg.lookup[Duration]("timeout")) { (endpoint, timeout) =>
+    def readKubernetesInfrastructure(kfg: KConfig): Option[Infrastructure.Kubernetes] =
+      (kfg.lookup[String]("endpoint") |@|
+       kfg.lookup[String]("version").flatMap(KubernetesVersion.fromString) |@|
+       kfg.lookup[String]("ca-cert").map(p => Paths.get(p)) |@|
+       kfg.lookup[String]("token") |@|
+       kfg.lookup[Duration]("timeout")
+      ) { (endpoint, version, caCert, token, timeout) =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/Config.scala
```diff
@@ -540,12 +540,16 @@ object Config {
         })
     }
 
-    def readKubernetesInfrastructure(kfg: KConfig): Option[Infrastructure.Kubernetes] = {
-      (kfg.lookup[String]("endpoint") |@| kfg.lookup[Duration]("timeout")) { (endpoint, timeout) =>
+    def readKubernetesInfrastructure(kfg: KConfig): Option[Infrastructure.Kubernetes] =
+      (kfg.lookup[String]("endpoint") |@|
+       kfg.lookup[String]("version").flatMap(KubernetesVersion.fromString) |@|
+       kfg.lookup[String]("caCert").map(p => Paths.get(p)) |@|
```
timperrett: we've avoided camel case for configuration paramaters, in favor of kebab case. Can we rename this to `ca-cert`?

core/src/main/scala/Config.scala
```diff
@@ -559,12 +563,8 @@ object Config {
         def checkServerTrusted(certs: Array[X509Certificate], authType: String): Unit = ()
       }
 
-    def getKubernetesPodCert(): Option[SSLContext] = {
-      // Auto-mounted at this path for pods
-      // https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
-      val certBundle = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-
-      val is = new FileInputStream(certBundle)
+    def getKubernetesPodCert(certPath: Path): Option[SSLContext] = {
```
timperrett: Just so I understand what's going on here, you're assuming that there's per-cluster PKI for interacting with the Kube API? Presumably this means the location that Nelson is deployed needs the certs on disk (this should be fine, just trying to grok it)

adelbertc: Yep that is the current assumption

core/src/main/scala/Config.scala
adelbertc: kebab-case, love it. will do

PR #69, opened 2018-03-31
Allow resource requests in manifest, fixes #66

### LINKED ISSUE #66: Add resource requests to manifest file
Currently Nelson's CPU and memory specs are resource limits, an upper bound on how many resources a deployment can be allocated. Similarly useful is a resource *request*, a lower bound on how many resources a deployment needs to function properly.

My current plan is to add this as an optional parameter to the current V1 manifest, defaulting it to be equal to the corresponding resource limit when not specified.

One question I have is during parsing/validation time we have some choices to make.. it is possible for the author to specify a resource **request** but not a resource **limt**. Currently resources limits have a default value (0.5 CPU, 512 MB memory), so I see two choices here:

1. Allow users to specify a request without a limit, and validate the specified request is less than the Nelson-provided default. This may make the code a bit trickier, I imagine we have to bubble up the default values [from the scheduler](https://github.com/getnelson/nelson/blob/master/core/src/main/scala/scheduler/NomadHttp.scala#L355-L356) to the validation logic so other parts of the validation can use it. Another option is to make an ADT that is either some symbolic default value that scheduler interpreters are free to interpret on their own, or a user-specified value.

2. Disallow users to specify a request without a limit - fail the validation if this happens.

@timperrett ?

adelbertc:
Prior to this change resource specs in the manifest file could only
specify limits, or upper bounds. This change adds support for
'cpu_request' and 'memory_request" which specifies requests, or a lower
bound - e.g. my app needs at least this much CPU/memory to function.

It seems Nomad does not support specifying lower bounds so in the Nomad
case the lower bound is ignored. Kubernetes (ostensibly) supports it and
so it uses the requests and limits appropriately.

Here is a quick breakdown of how resources are requested from Kubernetes
depending on what is provided in the manifest (note it is also designed
with backwards compatibility in mind):

- If a request is specified, a limit must be explicitly provided
- If request and limit are both specified, request must be <= limit
- If only limit is specified, request = limit (this is what Kubernetes
  does anyways at the time of implementation, but the impelmentation
  makes it explicit)
- If neither are specified, both request and limit are set to the same
  (and original) default value of 0.5 CPU and 512M memory


core/src/main/scala/KubernetesClient.scala
```diff
@@ -135,6 +135,25 @@ object KubernetesClient {
 }
 
 object KubernetesJson {
+  val DEFAULT_CPU = 0.5
```
timperrett: do we use uppercase constants elsewhere in the nelson codebase? don't think we do - this looks a little funny to my eye :-)

core/src/main/scala/Manifest.scala
```diff
@@ -69,8 +69,8 @@ object Manifest {
   }
 
   final case class Environment(
-    cpu: Option[Double] = None,
-    memory: Option[Double] = None,
+    cpu: Option[(Double, Double)] = None,
```
timperrett: I'm not wild about this tuple - can we either type alias `Double` to something semantically meaningful, or provide an ADT?

core/src/test/resources/nelson/manifest.v1.everything.yml
```diff
@@ -89,7 +89,9 @@ loadbalancers:
 # plans define `how` we want to deploy a unit, basically the runtime
 plans:
   - name: default-foobar
+    cpu_request: 0.25
```
timperrett: @kaiserpelagic what do you think about this naming?

core/src/main/scala/Manifest.scala
adelbertc: Leaning towards doing an ADT since that seems like the more typeful way to do things. I'm thinking something like:

```scala
sealed abstract class ResourceSpec[A] extends Product with Serializable {
  def fold[B](unspecified: B)(limitOnly: A => B)(bounded: (A, A) => B): B = ...
}

object ResourceSpec {
  // Hide the constructors since we want to make smart constructors for 'bounded'
  private class Unspecified[A]() extends ResourceSpec[A]
  private class LimitOnly[A](limit: A) extends ResourceSpec[A]
  private class Bounded[A](request: A, limit: A) extends ResourceSpec[A]

  def unspecified[A]: ResourceSpec[A] = new Unspecified[A]()

  def limitOnly[A](limit: A)(implicit A: Numeric[A]): Option[ResourceSpec[A]] =
    if (A.gteq(limit, A.zero)) Some(new LimitOnly(limit))
    else None

  def bounded[A](request: A, limit: A)(implicit A: Numeric[A]): Option[ResourceSpec[A]] =
    if (A.gteg(limit, request) && A.gteq(limit, A.zero)) Some(new Bounded(request, limit))
    else None
}
```

This would fit with us validating everything in the parser and from then on anything that sees `ResourceSpec` can be assured and patmat on if a resource is unspecified, limit only, or bounded, with appropriate checks. WDYT?

core/src/main/scala/KubernetesClient.scala
adelbertc: fixed!

core/src/main/scala/Manifest.scala
adelbertc: done!

timperrett:
@adelbertc this all looks great. Can you just update the documentation to include the changes to the manifest, and the behavior in the event that users do not specify the needful? Thanks for doing this.

adelbertc:
done!

adelbertc:
@timperrett ping? :)

adelbertc:
@timperrett Done, but can't merge without your approval :-)

adelbertc:
@timperrett I initially meant I needed your actual GitHub approval but apparently I can merge without it? 😆

timperrett:
@adelbertc i just contribute memes 🙌

PR #70, opened 2018-05-03
Remove remaining Tasks and cleanup some migrations

adelbertc:
- Change unsafeTimed to be safe using new IO cancellation feature
- Make Metrics.timer logic clearer
- Replace remaining uses of Task with IO

### DIFF
--- a/core/src/main/scala/Metrics.scala
+++ b/core/src/main/scala/Metrics.scala
@@ -17,7 +17,6 @@
 package nelson
 
 import cats.effect.IO
-import cats.syntax.applicativeError._
 import cats.syntax.apply._
 
 import io.prometheus.client._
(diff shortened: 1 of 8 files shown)


core/src/main/scala/Metrics.scala
```diff
@@ -168,7 +167,10 @@ object Metrics {
         _       <- onComplete.run(elapsed)
       } yield a
 
-      (io <* onSuccess).onError { case err => onFail(err) }
+      io.attempt.flatMap {
+        case Left(e)  => onFail(e) *> IO.raiseError(e)
+        case Right(a) => onSuccess *> IO.pure(a)
```
adelbertc: Based on the original use of [`Task#onFinish`](https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Task.scala#L50-L61) this seems to be a faithful translation, despite what the documentation of `onFinish` may suggest.

PR #71, opened 2018-05-10
s/scalaz.Free/cats.free.Free

adelbertc:
Part 2/N of Scalaz -> Cats migration, following #64 

- Replace `scalaz.Free` and `scalaz.Free.FreeC` with `cats.free.Free`
    - No longer need `Coyoneda` since `cats.free.Free` does the Coyoneda trick behind the scenes
- Remove `run*` helper methods since those were used mostly for a more readable `scalaz.Free.runFC`, but no longer needed since `cats.free.Free` does the Coyoneda trick already and `foldMap` works as expected
- Remove `ScalazHelpers` as all of them exist directly on the corresponding Cats data types, e.g. `inject` and `or`


timperrett:
@adelbertc can you put a description here explaining what you're doing?

core/src/test/scala/test/Interpreter.scala
```diff
@@ -114,7 +115,16 @@ sealed trait Interpreter[F[_], M[_], +S] {
...
 
-          case \/-(_) =>
+
+
+            // val pattern = patternF.apply[Free[F, A]]
+            // pattern.lift(sf) match {
+            //   case Some((it2, ma)) => ma.flatMap(loop(it2, _, Some(sf)))
+
+            //   case None => loop(fail, fc, lastSuspension)
+            // }
```
adelbertc: oops

adelbertc:
@timperrett  Updated description.

Re: CI failure..
After staring at the CI errors for an hour I think I found the bug, it's a funny one. The NPE originates in [this stub](https://github.com/adelbertc/nelson/blob/26faa1ed8fae181215e206091379c2c8c62bba2a/core/src/test/scala/RoutingFixtures.scala#L146) - usually the `null` stubbing gets replaced (via case class copy) before it actually gets used. However, due to the following change in several of the tests:

```diff
-    nelson.storage.run(config.storage, insertFixtures(testName)).unsafeRunSync()
+    insertFixtures(testName).foldMap(config.storage).unsafeRunSync()
```

Note that in the former `config.storage` is evaluated before `insertFixtures(testName)` is whereas in the latter it is flipped. The call to `insertFixtures` eventually ends up hitting the line mentioned above, causing it to NPE. HOWEVER, part of the [construction of `config.storage`](https://github.com/adelbertc/nelson/blob/26faa1ed8fae181215e206091379c2c8c62bba2a/core/src/test/scala/NelsonSuite.scala#L188-L193) involves replacing the aforementioned `null` stubbing so in the former case by the time it got to `insertFixtures` everything is OK, but in the latter case `insertFixtures` is evaluated first and therefore there was no chance to replace the stubs. Thus, NPE.

adelbertc:
Tested on a Kubernetes cluster, released a job, verified it deployed and ran, released a new version, verified old job was GC'd and expiration times worked as expected - ready for review!

core/src/main/scala/workflows/Magnetar.scala
```diff
@@ -69,14 +68,14 @@ object Magnetar extends Workflow[Unit] {
...
-    logToFile(d.id, s"removing policy from vault: ${vaultLoggingFields(sn, ns = ns.name, dcName = dc.name)}") >>
-    deletePolicyFromVault(d.stackName, ns.name) >>
-    logToFile(d.id, s"removing alerts from consul ${alerts.alertingKey(sn)}") >>
-    deleteAlertsFromConsul(d.stackName) >>
-    logToFile(d.id, s"removing discovery tables from consul ${routing.Discovery.consulDiscoveryKey(sn)}") >>
-    deleteDiscoveryInfoFromConsul(sn) >>
-    logToFile(d.id, s"instructing ${dc.name}'s scheduler to decommission ${sn}") >>
-    delete(dc,d) >>
+    logToFile(d.id, s"removing policy from vault: ${vaultLoggingFields(sn, ns = ns.name, dcName = dc.name)}") *>
+    deletePolicyFromVault(d.stackName, ns.name) *>
```
ChristopherDavenport: Confirming this didn't rely on lazily building the `WorkflowF` as construction here is strict.

adelbertc: It's all `Free` so should be fine

PR #75, opened 2018-05-27
Finish migration from Scalaz to Cats

adelbertc:
- Migrate remaining `NonEmptyList` and `Validation`
- Migrate `\/` and `DisjunctionT`
- Clean up `CatsHelpers`
- Migrate `OptionT`
- Migrate `==>>` to `SortedMap` and migrate `Order`
    - To test `Order` laws had to bring in `cats-laws`.. which requires bringing in a new version of ScalaCheck.. which requires bringing in a new version of ScalaTest (yay dependencies)
- Migrate `Kleisli`
- Migrate `RWST`
- Migrate `Interpreter`
- Inline/port `scalaz.{@@, Tag}` to Cats
- Migrate Quiver
- Bump `http4s-argonaut61` to completely remove Scalaz on classpath


adelbertc:
Did the usual basic test on K8s - released a job, verified it deployed and ran, released a new version, verified old job was GC'd and expiration times worked as expected. Ready to be reviewed!

core/src/main/scala/Datacenter.scala
```diff
@@ -192,8 +187,7 @@ object Datacenter {
 
   object Namespace {
     implicit def namespaceOrder: Order[Namespace] =
-      Order[String].contramap[Namespace](_.datacenter) |+|
-      Order[String].contramap[Namespace](_.name.asString)
+      Order.whenEqual(Order.by(_.datacenter), Order.by(_.name.asString))
```
kaiserpelagic: I'm not familiar with `whenEqual` but assuming the sematics remain the same.

core/src/main/scala/BiggerApplies.scala
```diff
@@ -1,34 +0,0 @@
...
-//:   See the License for the specific language governing permissions and
-//:   limitations under the License.
-//:
-//: ----------------------------------------------------------------------------
-package nelson
-
-import scalaz._
-
-// Triskaidekaphobia
-object BiggerApplies {
```
kaiserpelagic: It was a sad day when I had to write this (or was that Ross, I can't remember), but glad it see it go. Assuming you are using `mapN` in places where this would have been applicable.

core/src/main/scala/package.scala
```diff
@@ -57,6 +55,12 @@ package object nelson {
   type DNSName = String
   type DeploymentStatusString = String
 
+/** Copied and adapted from Scalaz's Tag implementation.
+  * https://github.com/scalaz/scalaz/blob/v7.1.17/core/src/main/scala/scalaz/package.scala
+  */
+  private[this] type Tagged[A, T] = { type Tag = T; type Self = A; }
```
kaiserpelagic: I'm curious why this doesn't exist in `cats`. Is it something that should be added?

kaiserpelagic:
@adelbertc where you able to test `services` on k8s, or just jobs? We have a decent amount of unit tests around the routing stuff so I feel pretty comfortable if you weren't able to test that. 

Anyways this LGTM, seems to be pretty mechanical. Also, thanks for taking this on.

core/src/main/scala/BiggerApplies.scala
adelbertc: Yessir, I always wondered if they bothered to go up to 13, why not just go up the rest of the way to 22?

core/src/main/scala/package.scala
adelbertc: There's been a couple competing implementations and nobody has converged on something yet, I think at this point a lot of folks are waiting for [SIP-35 opaque types](https://docs.scala-lang.org/sips/opaque-types.html) to land.

adelbertc:
@kaiserpelagic Just jobs, unfortunately routing is currently not implemented in the K8s backend yet (SHHHH), but now that the Cats migration is done I want to move on to doing that pretty soon. @timperrett and I have talked a couple of times of how that might look, essentially we're going to co-opt the K8s Service and Labels mechanisms.

adelbertc:
All aboard the [catboat](https://en.wikipedia.org/wiki/Catboat)! Or if you prefer, the [catamaran](https://en.wikipedia.org/wiki/Catamaran).

PR #78, opened 2018-08-06
More Kubernetes features and additions

adelbertc:
Initially started as just implementing volume mounts, but added some other stuff too..

- Add support for empty "scratch" directory volume mounts
- Add support for K8s 1.9.x and 1.10.x
- Allow operators to choose between in or out of cluster - if deploying in-cluster it will assuming the appropriate RBAC is configured and use Pod-mounted certs
- Build image to run as non-root
- Stub Consul client in K8s mode so we don't get bogus Consul errors in logs

### DIFF
--- a/core/src/main/scala/Config.scala
+++ b/core/src/main/scala/Config.scala
@@ -620,25 +620,6 @@ object Config {
           kfg.lookup[String](s"proxy-credentials.password")
         ).mapN((a,b) => Infrastructure.ProxyCredentials(a,b))
 
-      val consul = {
-        val a = kfg.require[String]("infrastructure.consul.endpoint")
-        val b = kfg.require[Duration]("infrastructure.consul.timeout")
-        val c = kfg.lookup[String]("infrastructure.consul.acl-token")
-        val d = kfg.lookup[String]("infrastructure.consul.username")
-        val e = kfg.lookup[String]("infrastructure.consul.password")
-        val client = http4sClient(b, 20)
-        val http4sConsul = (d,e) match {
-          case (None,None) => client.map(consulClient => Http4sConsul.client(Infrastructure.Consul(new URI(a), b, c, None), consulClient))
-          case (Some(u),Some(pw)) =>
-            client.map(consulClient => Http4sConsul.client(Infrastructure.Consul(new URI(a), b, c,
-              Some(Infrastructure.Credentials(u,pw))), consulClient))
-          case _ =>
-            log.error("If you configure the datacenter to have a consul username, or consul password, it must have both.")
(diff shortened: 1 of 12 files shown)


core/src/main/scala/Config.scala
```diff
@@ -654,29 +638,49 @@ object Config {
...
             consulClient  <- consul
             sched         <- readNomadScheduler(schedConfig.subconfig("nomad"))
             healthChecker = health.Http4sConsulHealthClient(consulClient)
-          } yield (sched, healthChecker)
+          } yield (sched, healthChecker, consulClient)
+
         case Some("kubernetes") =>
           readKubernetesClient(schedConfig.subconfig("kubernetes")).map { client =>
-            (new scheduler.KubernetesHttp(client), health.KubernetesHealthClient(client))
+            (new scheduler.KubernetesHttp(client), health.KubernetesHealthClient(client), StubbedConsulClient)
```
timperrett: `StubbedConsulClient`? Would be good to understand the rationale here.

adelbertc: I confess this was selfishly motivated - we are not using Consul with Kubernetes not are we (or you afaik) planning on it, but in retrospect it is totally possible (I think [Stripe](https://stripe.com/blog/operating-kubernetes) maybe does this?).

Anyways assuming that is the case, there is no sane values to put for the Consul portion of the config (which we need to make more generic BTW) and so when Nelson goes to refresh routing tables you get this nice stack trace in the logs that is bogus.

Unsure what to do here:

- The Correct Way™ would be to give an option in the config to signal what backend to use for the routing layer, similar to what we do for the scheduler
- Try to make a dummy call to Consul and if that fails fallback to the stubbed client

This of course is a stopgap while we're in the "Kubernetes without Consul currently has no routing support from Nelson" stage. Curious to hear your thoughts

core/src/test/resources/nelson/manifest.v1.everything.yml
```diff
@@ -108,6 +108,11 @@ plans:
       - ref: s3
         uri: http://s3.aws.com
     expiration_policy: retain-latest-two-major
+    volumes:
+      - name: an-empty-dir
+        mountPath: /foo/bar
+        emptyDir:
```
timperrett: I'm not sure we should be specific about the kind of volume here based on our discussion at lunch. If we go for templating later, then we can simply have the user specify the size of the volume they'd like. WDYT?

timperrett:
Other than that one comment, I think this looks fine 😄

adelbertc:
@timperrett Made the change, OK to merge?

PR #82, opened 2018-08-14
Add Slipway section to User Guide, fixes #80

### LINKED ISSUE #80: Docs on Slipway

adelbertc:


timperrett:
This is great, but misses one key design choice: deployables bridge the gap between logical definition and physical definition. With continuous delivery you're always producing new versions, so this design avoids tedious management of the versions specified in scheduler configs, whilst avoiding the need for a mutable tag like `latest` etc.

PR #92, opened 2018-08-19
blueprints/api

timperrett:
resolves #84 and #87

Issue #86, opened 2018-08-17
Major revision of the manifest parser and linter so that workflow is part of `plan`


timperrett:
- [ ] Add a version to the manifest
- [ ] update the parser so that we try to parse version first and then select a parser, but fall back to assuming V1 of the parser.

Proposed solution for manifest versioning:
`version: stable/v1` <- present master
`version: stable/v2` <- post-blueprint
`version: unstable/v3` <- master

Issue #61, opened 2018-01-13
Docs for Nelson on K8s


adelbertc:
Now that initial K8s integration is merged 🔥 🔥 🔥 we should probably document the current state, what is/isn't supported, etc.

I'll try to find some time soon to do this, want to do this before I make noise on Twitter, etc. about it. What might be the best place to put this? The `kubernetes-configuration` repo? README? Site? It's very preliminary integration so I'm hesitant to put it somewhere as central as the site, but maybe README?

timperrett:
Probally putting some mentions here would be useful: https://getnelson.github.io/nelson/#installation-configuration

but broadly yes, this is something that absolutely should be done :-)

PR #93, opened 2018-08-19
blueprints/templating

adelbertc:
- [X] Split out deployment and deletion from `KubernetesClient`
- [X] Prep schedulers to delegate relevant work to blueprints
- [X] Read template and blueprint from DB
- [X] Translate manifest to template rendering environment
- [X] Make scheduler backends actually use blueprints
    - [x] Kubernetes
        - [X] Launch
        - [x] Delete
- [X] Implement shelling out (to `kubectl`)
- [X] Drop Nomad support
- [x] Re-implement K8s backend functionality with `kubectl`
    - [x] scheduler summary
    - [x] health check
- [x] Delete hard-coded sidecars
- [ ] Test all this with a live cluster


core/build.sbt
```diff
@@ -43,6 +44,7 @@ libraryDependencies ++= Seq(
   "org.tpolecat"               %% "doobie-hikari"                      % V.doobie,
   "org.spire-math"             %% "spire"                              % "0.11.0",
   "org.flywaydb"                % "flyway-core"                        % "3.2.1",
+  "io.circe"                   %% "circe-yaml"                         % "0.8.0",
```
timperrett: Are we using this dep? I don't notice its usage lower in the changeset

core/build.sbt
```diff
@@ -67,8 +69,6 @@ addCompilerPlugin(dependencies.macroparadise.plugin)
 
 initialCommands in console := """
 import nelson._
```
timperrett: I fixed this in my PR :-) https://github.com/getnelson/nelson/pull/92/files#diff-10e6c0c6d9998a5d5b6e19e45f1485abR71

core/src/main/scala/scheduler/Kubectl.scala
```diff
@@ -0,0 +1,12 @@
+package nelson
+package scheduler
+
+import cats.effect.IO
+
+object Kubectl {
```
timperrett: Might be work a bit of documentation here. How do these end up becoming `sys.process`?

core/src/test/resources/nelson/canopus_cronJob.mustache
```diff
@@ -0,0 +1,56 @@
+apiVersion: batch/v1beta1
```
timperrett: Naming of this file - can we be consistent between snake case and camel case? (snake case prefered here)

core/src/main/scala/workflows/Canopus.scala
```diff
@@ -37,7 +37,7 @@ object Canopus extends Workflow[Unit] {
       i <- DockerOp.extract(Versioned.unwrap(vunit)).inject[WorkflowOp]
       _ <- status(id, Pending, "Canopus workflow about to start")
       _ <- logToFile(id, s"Instructing ${dc.name}'s scheduler to handle service container")
-      l <- launch(i, dc, ns.name, vunit, p, hash)
+      l <- launch(i, dc, ns.name, vunit, p, None, hash)
```
timperrett: So somehow here we need to connect the dots to get the blueprint passed in, right?

core/src/main/scala/blueprint/DefaultBlueprints.scala
```diff
@@ -0,0 +1,26 @@
+package nelson
+package blueprint
+
+import cats.effect.IO
+
+import fs2.{io, text}
+
+object DefaultBlueprints {
```
timperrett: This is fantastic.

timperrett:
In general I think this is totally heading in the right direction and that we're really close. A couple things come to mind reviewing this:

- All references to the logging sidecar that nelson presently knows about should be removed
- Using the blueprints should significantly simplify the configuration for each scheduler
- I just noticed we did not even document the k8s support, so we should file a ticket for documenting blueprints extensively.

core/src/main/scala/scheduler/Kubectl.scala
adelbertc: Are you asking how we're actually going to shell out here? If so I am entertaining two options atm, both from the venerable @tpolecat : 

- FS2 approach: https://gist.github.com/tpolecat/f7fac786e084d8e0e2f642b1e8646513
- vanilla IO approach: https://github.com/tpolecat/ocs3/blob/develop/modules/ctl/src/main/scala/low/io.scala

core/src/test/resources/nelson/canopus_cronJob.mustache
adelbertc: Ah yes good call :-) It was originally `canopus_cronjob` and I renamed it on a whim completely forgetting it used to at least have a naming convention

core/src/main/scala/scheduler/Kubectl.scala
timperrett: I'd probally go for the vanilla I/O approach as its more obvious - we got into trouble doing fancy I/O stuff with streams in funnel, as it was far harder to maintain and understand semantics.

core/src/main/scala/workflows/Canopus.scala
adelbertc: I imagine this will be implemented as part of https://github.com/getnelson/nelson/issues/86

adelbertc:
* Logging sidecar dropped
* Because we decided to just shell out for everything including scheduler summaries and health checks, we can delete whole swaths of the K8s scheduler code. The same would hold for Nomad but per our earlier convo we're momentarily dropping Nomad support 
* As for K8s docs, a ticket exists https://github.com/getnelson/nelson/issues/61 - definitely going to want to have that done before I give my SBTB talk 😅

core/src/main/scala/blueprint/Render.scala
```diff
@@ -0,0 +1,127 @@
+package nelson
+package blueprint
+
+import nelson.Datacenter.StackName
+import nelson.Manifest._
+import nelson.blueprint.EnvValue._
+import nelson.docker.Docker.Image
+
+object Render {
```
timperrett: This file should be well documented as it's the core of the blueprint templaing engine

adelbertc: Yep, will follow up all the docs and tests in future PR as we discussed.

PR #100, opened 2018-08-31
The 0.10.x series should publish 0.10.x artifacts

adelbertc:


adelbertc:
@timperrett Do we need to do anything else to get CI/CD setup for 0.10.x?

.buildkite/pipeline.yml
```diff
@@ -5,7 +5,7 @@ steps:
     concurrency: 1
     concurrency_group: primary
     timeout_in_minutes: 45
-    branches: master
+    branches: "master 0.10.x"
```
timperrett: Is this the correct syntax? I thought it was an array, not space delimited

adelbertc: I was surprised to, but according to https://buildkite.com/docs/pipelines/branch-configuration#branch-pattern-examples it is a space separated string

timperrett:
The only thing that is needed to build this is to have the CI droplet running, which i can do tomorrow night when I get access to a computer

adelbertc:
@timperrett Sweet! One more thing - I assume the versions that are actually published use the MAJOR.MINOR from `version.sbt` but the patch just uses `$BUILD_NUMBER` or whatever. In that case is the `$BUILD_NUMBER` shared across branches? Or is it going to start trying to publish from `0.10.{0, 1, ...}`? If it starts over we're going to hit some duplicate artifact versions since at the very least we have an existing `0.10.63` or something.

timperrett:
It’s monotonic for the build pipeline, so it’s unique across scm sources, so there shouldn’t be any issues at all.

adelbertc:
@timperrett OK so the next build of 0.10.x branch will probably attempt to publish 0.10.64+ or something ?

timperrett:
Yep, the numbers won’t be sequential of course, but greater numbers will always trump lesser numbers

PR #111, opened 2018-09-08
Kubernetes/jwt

timperrett:
This adds support for the Vault kubernetes auth mechinism:
https://www.vaultproject.io/api/auth/kubernetes/index.html

Here, Nelson is simply taking the policies it provisioned per-stack, and enabling vault to use the stack-specific JWT to authenticate against Vault. In order to do this, the following administrator steps are required:

1. Create a service account in kubernetes that will be used for token validation 

```
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: vault-auth
  namespace: default
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: role-tokenreview-binding
  namespace: default
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:auth-delegator
subjects:
- kind: ServiceAccount
  name: vault-auth
  namespace: default
```

2. submit this to kubernetes:

```
kubectl apply -f /path/to/above.yml
```
3. Next configure Vault to have the auth backend:

```
# fetch the JWT token
stack_name=vault-auth
token_name=$(kubectl -n default get serviceaccount vault-auth -o jsonpath='{.secrets[0].name}')
token_data=$(kubectl get secret "${token_name}" -o jsonpath='{.data.token}' | base64 --decode)

# setup a principle for this JWT
vault write auth/kubernetes/config \
token_reviewer_jwt="${token_data}" \
kubernetes_host=$KUBEURL \
kubernetes_ca_cert=@/path/to/ca
```

4. At runtime, from within the deployed pod (an application deployed with Nelson), do something like this in your container (or init container):

```
# fetch the JWT token
token_data=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)

vault write -format=json -address="${VAULT_ADDR}" \
"auth/kubernetes/login" \
role="${NELSON_STACK}" \
jwt="${token_data}"
```

### DIFF
--- a/core/src/main/scala/workflows/Pulsar.scala
+++ b/core/src/main/scala/workflows/Pulsar.scala
@@ -17,6 +17,10 @@ object Pulsar extends Workflow[Unit] {
   val name: WorkflowRef = "pulsar"
 
   def deploy(id: ID, hash: String, vunit: UnitDef @@ Versioned, p: Plan, dc: Datacenter, ns: ManifestNamespace): WorkflowF[Unit] = {
+    val unit = Manifest.Versioned.unwrap(vunit)
+    val sn = Datacenter.StackName(unit.name, vunit.version, hash)
+    val rs = unit.dependencies.keys.toSet ++ unit.resources.map(_.name)
+
     // When the workflow is completed, we typically want to set the deployment to "Warming", so that once
     // consul indicates the deployment to be passing the health check, we can promote to "Ready" (via the
     // DeploymentMonitor background process).  However, units without ports are not registered in consul, and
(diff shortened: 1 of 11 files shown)


core/src/main/scala/workflows/Pulsar.scala
```diff
@@ -29,6 +33,12 @@ object Pulsar extends Workflow[Unit] {
     for {
       i <- DockerOp.extract(Versioned.unwrap(vunit)).inject[WorkflowOp]
       _ <- status(id, Pending, "Pulsar workflow about to start")
+      //// write the policies to vault
+      _  <- logToFile(id, s"writing policy to vault: ${vaultLoggingFields(sn, ns = ns.name, dcName = dc.name)}")
+      _  <- writePolicyToVault(cfg = dc.policy, sn = sn, ns = ns.name, rs = rs)
+      //// create a vault k8s auth role
+      _  <- logToFile(id, s"writing k8s auth role to vault...")
```
adelbertc: Spell out Kubernetes and capitalize 'v' in 'vault'? :)

core/src/main/scala/workflows/Pulsar.scala
```diff
@@ -29,6 +33,12 @@ object Pulsar extends Workflow[Unit] {
...
+      // _  <- writeKubernetesRoleToVault(dc = dc, sn = sn, ns = ns.name)
```
adelbertc: Is this meant to be commented out?

adelbertc:
We'll want to add what you wrote in the description to the actual docs

Issue #90, opened 2018-08-18
Migrate the slack integration to Http4s 


timperrett:
The slack client is the last place we use n8han's dispatch library; we should migrate it to http4s to be consistent across the rest of the code base.

timperrett:
Awesome. This is so overdue it’s unreal!

PR #113, opened 2018-09-08
Dispatch be gone

adelbertc:
Closes https://github.com/getnelson/nelson/issues/90


adelbertc:
Do not merge yet, looks like there are some bugs talking to GitHub.

core/src/main/scala/Config.scala
```diff
@@ -375,24 +385,21 @@ object Config {
...
+    val timeout = cfg.require[FiniteDuration]("nelson.timeout")
+    val http = httpBuilder(BlazeClientConfig.defaultConfig.copy(requestTimeout = timeout))
 
     val pools = Pools.default
 
     val nomadcfg = readNomad(cfg.subconfig("nelson.nomad"))
 
     val gitcfg = readGithub(cfg.subconfig("nelson.github"))
-    val git = new Github.GithubHttp(gitcfg, http0)
+    val git = http.map(client => new Github.GithubHttp(gitcfg, client, timeout, pools.defaultExecutor))
```
adelbertc: Reminder to my future self debugging this - I believe this will create a new HTTP client each time `git` (or `slack`) is used which may explain the slowdown and number of connections I see being created in the logs. Should probably/definitely fix this.

adelbertc:
@timperrett Ready to go down, tested nelson login, repo sync, repo enable, and webhook, all good ✅

PR #115, opened 2018-09-10
Totally Redesign the Documentation Website

timperrett:
This is not quite ready yet, but here's the WIP:

screen shot 2018-09-10 at 20 37 15
screen shot 2018-09-11 at 07 35 40
screen shot 2018-09-11 at 07 35 45

screen shot 2018-09-10 at 20 44 42


docs/src/hugo/layouts/index.html
```diff
@@ -0,0 +1,146 @@
...
+        </div>
+      </div> <!-- /jumbotron -->
+
+      <div class="jumbotron jumbosplash bg-blue-5">
+        <div class="container">
+          <div class="row">
+            <div class="col-md-12">
+              <h2 class="semibold">Open-source, Container-first Orchestration</h2>
+              <p class="lead">
+                Nelson is a cutting edge orchestration engine that empowers users to focus on building their products, whilst providing operators with the safety and confidence user applications are meeting the organizationl best practices. Having its roots in type-theory and academic research, Nelson fuses field experience with formal rigor to make deployments something that "just works".
```
adelbertc: As much as I am amused at the mentions, I feel like putting "type theory" and "academic research" might scare some people off, at least on the splash page. Maybe we can put something vaguer or more general, or just even drop the first chunk and start with "Nelson fuses field experience with formal rigor.."

Also maybe replace "users" which can be vague (like end-users?) with "service owners" or "product owners."

docs/src/hugo/layouts/index.html
```diff
@@ -0,0 +1,146 @@
...
+              </p>
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-md-6">
+              <h2>Immutable Deployments</h2>
+              <p>Nelson takes immutable infrastructure to the next level: each and every deployment is fixed for its lifetime, and upgrades are conducted by migrating traffic between versions. This enables powerful production-testing and experimentation.</p>
```
adelbertc: Could be clearer - maybe

> Nelson is built from the ground-up to support immutable deployments: every deployment and dependency is explicitly and semantically versioned, allowing Nelson to have full knowledge of the deployment graph and enabling easy traffic shifting, production testing, and experimentation.

docs/src/hugo/layouts/index.html
```diff
@@ -0,0 +1,146 @@
...
+              <h2>Scheduler Agnostic</h2>
+              <p>Users face a plethora of choices when it comes to picking a scheduler - you may even be running multiple kinds of scheduler. Nelson enables operators to focus on scheduling (without lossing power), whilst users focus on shipping their product.</p>
+            </div>
+            <div class="col-md-6" style="padding-top: 2em;">
+              <h2>Git-centric</h2>
+              <p>Out of the box integration with Github (SaaS and Enterprise versions) means Nelson already integrates into the workflow you know and love. Simply create a project release on Github and Nelson will distribute and run your program container.</p>
+            </div>
+            <div class="col-md-6" style="padding-top: 2em;">
+              <h2>Secure By Default</h2>
+              <p>The easy thing should also be the safe thing. Nelson has security baked-in: every stack launched with Nelson has unique Vault policies, with ephemeral, rotating credentials that are unique per-container.</p>
```
adelbertc: Perhaps a weaker statement in case folks coming across this page aren't/don't want to run Vault. Maybe something like "Security Integrations" and "Nelson supports Vault policies.." as opposed to it being explicitly baked -in.

docs/src/hugo/layouts/index.html
```diff
@@ -0,0 +1,146 @@
...
+        </section>
+      </div>
+
+      <div class="jumbotron jumbosplash bg-blue-5">
+        <div class="container">
+          <div class="row">
+
+            <div class="col-md-4">
+              <h2>Topological Graph</h2>
+              <p>Distributed systems are fundimentally graphs. Nelson embraces this and models your entire system topology as a graph, allowing the system to know which systems interact, and what exposure any given system represents.</p>
```
adelbertc: fundamentally*

PR #122, opened 2018-09-15
Fix OAuth login

timperrett:


core/src/main/scala/Config.scala
```diff
@@ -77,7 +77,7 @@ final case class GithubConfig(
   val loginEndpoint = {
     val queryParams = Map(
       ("client_id", List(clientId)),
-      ("redirect_uri", List(encodeURI(redirectUri))),
+      ("redirect_uri", List(redirectUri)),
```
timperrett: @adelbertc this is the reason it wasn't working: it gets double encoded so github was saying the URLs did not match.

PR #124, opened 2018-09-17
blueprints/manifest

### LINKED ISSUE #86: Major revision of the manifest parser and linter so that workflow is part of `plan`

timperrett:
resolves #86 

This PR adds support for validating manifests with blueprints. Note that the implementation is done in two steps, and is a breaking change for manifests working with Nelson 0.11.x or older (essentially all manifests).

1. The blueprint reference (e.g. `something@HEAD`) is first checked for structural validity in the parser; in this case three forms for blueprint reference are allow, as denoted by the blueprint reference: https://github.com/getnelson/nelson/blob/master/core/src/main/scala/blueprint/Blueprint.scala#L39

2. Assuming the manifest is parsable, then the `ManifestValidator` will then be engaged to check the blueprint reference is actually valid and present in the database. In the event the blueprint is not found, the process will short-circuit with an error. In the event the manifest validator finds no errors, it will update the supplied manifest with the plans containing the hydrated `Blueprint`'s

### DIFF
--- a/core/src/main/scala/Manifest.scala
+++ b/core/src/main/scala/Manifest.scala
@@ -20,6 +20,7 @@ import nelson.Manifest._
 import nelson.cleanup.ExpirationPolicy
 import nelson.notifications.NotificationSubscriptions
 import nelson.storage.StoreOp
+import nelson.blueprint.Blueprint
 
 import cats.{~>, Foldable, Monoid}
 import cats.data.{Kleisli, NonEmptyList, ValidatedNel}
@@ -116,7 +117,9 @@ object Manifest {
     policy: Option[ExpirationPolicy] = None,
     trafficShift: Option[TrafficShift] = None,
     volumes: List[Volume] = List.empty,
-    ephemeralDisk: Option[Int] = None
+    ephemeralDisk: Option[Int] = None,
+    workflow: Option[Workflow[Unit]] = None,
+    blueprint: Option[Either[BlueprintRef, Blueprint]] = None
   )
 
   final case class Namespace(
--- a/core/src/main/scala/Manifest.scala
+++ b/core/src/main/scala/Manifest.scala
@@ -53,16 +53,9 @@ object Manifest {
     dependencies: Map[String, FeatureVersion],
     resources: Set[Resource],
     alerting: Alerting,
-    workflow: Workflow[Unit],
     ports: Option[Ports],
     deployable: Option[Deployable],
-    meta: Set[String],
-    // maintained for backwards compatibility (manifest).
-    // these fields are also defined in the plan.
-    // if they are absent in the plan then fallback to
-    // what's defined here, otherwise use a sensible default
-    schedule: Option[Schedule] = None,
-    policy: Option[ExpirationPolicy] = None
+    meta: Set[String]
   )
 
   final case class Plan(
@@ -71,7 +64,7 @@ object Manifest {
   )
 
   object Plan {
-    val default = Plan("default", Environment())
+    val default = Plan("default", Environment(workflow = Magnetar))
   }
 
   sealed abstract class ResourceSpec extends Product with Serializable {
@@ -118,7 +111,7 @@ object Manifest {
     trafficShift: Option[TrafficShift] = None,
     volumes: List[Volume] = List.empty,
     ephemeralDisk: Option[Int] = None,
-    workflow: Option[Workflow[Unit]] = None,
+    workflow: Workflow[Unit] = Magnetar,
     blueprint: Option[Either[BlueprintRef, Blueprint]] = None
   )
 
@@ -286,13 +279,13 @@ object Manifest {
   )
 
   def isPeriodic(unit: UnitDef, plan: Plan): Boolean =
-    unit.schedule.isDefined || plan.environment.schedule.isDefined
+    plan.environment.schedule.isDefined
 
   def getSchedule(unit: UnitDef, plan: Plan): Option[Schedule] =
-    plan.environment.schedule orElse unit.schedule
+    plan.environment.schedule
 
   def getExpirationPolicy(unit: UnitDef, plan: Plan): Option[cleanup.ExpirationPolicy] =
-    plan.environment.policy orElse unit.policy
+    plan.environment.policy
 
   def toAction[A](a: A, dc: Datacenter, ns: Namespace, p: Plan, n: NotificationSubscriptions)(implicit A: Actionable[A]): Action = {
     val hash = randomAlphaNumeric(desiredLength = 8) // create a unique hash for this deployment
--- a/core/src/main/scala/Manifest.scala
+++ b/core/src/main/scala/Manifest.scala
@@ -64,7 +64,7 @@ object Manifest {
   )
 
   object Plan {
-    val default = Plan("default", Environment(workflow = Magnetar))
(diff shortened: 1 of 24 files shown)


core/src/main/scala/Manifest.scala
```diff
@@ -70,7 +64,7 @@ object Manifest {
   )
 
   object Plan {
-    val default = Plan("default", Environment())
+    val default = Plan("default", Environment(workflow = Magnetar))
```
adelbertc: If anything we'd want it to be Pulsar or Canopus now right? :)

core/src/main/scala/Manifest.scala
```diff
@@ -116,7 +110,9 @@ object Manifest {
     policy: Option[ExpirationPolicy] = None,
     trafficShift: Option[TrafficShift] = None,
     volumes: List[Volume] = List.empty,
-    ephemeralDisk: Option[Int] = None
+    ephemeralDisk: Option[Int] = None,
```
adelbertc: Given we have `volumes` now do we still want to keep `ephemeralDisk`? Prob not this PR but a future one

core/src/main/scala/ManifestValidator.scala
```diff
@@ -208,33 +208,65 @@ object ManifestValidator {
...
           .sequence
           .map(l => Foldable[List].fold(l))
       }
     }
 
+    def validateBlueprints(m: Manifest): IO[(Valid[Unit], List[Plan])] = {
+      val foooo: IO[List[Valid[Plan]]] = m.plans.foldLeft(IO.pure(List.empty[Valid[Plan]])) { (res, plan) =>
+        val result: IO[Valid[Plan]] = plan.environment.blueprint match {
+          // supplied manifest held a manifest reference
+          case Some(Left((ref,revision))) => {
```
adelbertc: For the sake of type safety do we want to try the type parameter trick here? Where we have like:

```scala
case class Plan[A](...)
```

and set the type parameter to like `Unit` if its unsaturated and `Blueprint` if it is saturated or something? That way we can ensure in the scheduler we never get an unsaturated `Blueprint`

core/src/main/scala/ManifestValidator.scala
```diff
@@ -208,33 +208,65 @@ object ManifestValidator {
...
+          // executes
+          case None            => IO.pure(plan.validNel)
+        }
+        (result, res).mapN(_ :: _)
+      }
+
+      for {
+        a <- foooo
+        b  = Foldable[List].fold(a.map(_.map(_ => ())))
+        c  = a.traverse(_.toList).flatMap(identity)
```
adelbertc: This can be `a.flatTraverse(_.toList)`

core/src/main/scala/ManifestValidator.scala
```diff
@@ -208,33 +208,65 @@ object ManifestValidator {
...
+        d <- IO.pure((b, c))
+      } yield d
```
adelbertc: Just do `yield (b, c)` ?

core/src/main/scala/scheduler/KubernetesShell.scala
```diff
@@ -70,7 +70,15 @@ final class KubernetesShell(
       }
     }
 
-    val template = blueprint.fold(fallback)(IO.pure)
+    // NOTE: by this point in the system, we know we're dealing with
+    // a hydrated blueprint (i.e. passes manifest validation and exists
+    // in the database) so we simply take the supplied plan and extract
+    // the `Blueprint`, and `Template` in turn.
+    val template = plan.environment.blueprint
+      .flatMap(_.right.toOption)
```
adelbertc: This can have hidden/strange behavior if we have an internal bug and provide an unsaturated (hydrated?) blueprint and silently fallback to a default - see type parameter trick above.

core/src/main/scala/scheduler/NomadHttp.scala
```diff
@@ -132,8 +131,11 @@ final class NomadHttp(
   private def buildName(sn: StackName): String =
     cfg.applicationPrefix.map(prefix => s"${prefix}-${sn.toString}").getOrElse(sn.toString)
 
-  private def launch(u: UnitDef, hash: String, version: Version, img: Image, dc: Datacenter, ns: NamespaceName, plan: Plan, blueprint: Option[Template]): IO[String] =
-    blueprint.fold(launchDefault(u, hash, version, img, dc, ns, plan)) { template =>
+  private def launch(u: UnitDef, hash: String, version: Version, img: Image, dc: Datacenter, ns: NamespaceName, plan: Plan): IO[String] =
+    plan.environment.blueprint
+      .flatMap(_.right.toOption)
```
adelbertc: See my above comment on silent bugs here

core/src/main/scala/yaml/ManifestV1Parser.scala
```diff
@@ -179,7 +181,16 @@ object ManifestV1Parser {
...
+    // at this point, we do not have access to the database so we can only
+    // check that the specified blueprint is syntactically valid. The manifest
+    // validator will hydrate the reference into a Blueprint proper.
+    val validateBlueprint =
+      Option(raw.workflow).traverse(x =>
+        Blueprint.parseNamedRevision(x.blueprint)
+          .fold(_ => None, x => Some(Left(x)))
+            .toValidNel(YamlError.invalidBlueprintReference(x.blueprint)))
+
+    Apply[YamlValidation].map17(
```
adelbertc: We're getting close to the limit 😆 In the Scalaz version we used to use we would've had to dance around at this point

core/src/main/scala/yaml/ManifestV1Parser.scala
```diff
@@ -507,6 +514,12 @@ class PlanYaml {
   @BeanProperty var traffic_shift: TrafficShiftYaml = _
   @BeanProperty var volumes: JList[VolumeYaml] = new java.util.ArrayList
   @BeanProperty var ephemeral_disk: Int = -1
+  @BeanProperty var workflow: WorkflowYaml = _
+}
+
+class WorkflowYaml {
+  @BeanProperty var kind: String = "magnetar"
```
adelbertc: Do we want magnetar as the magic default?

core/src/main/scala/ManifestValidator.scala
timperrett: It won't have it:
```

[error] /ManifestValidator.scala:258:43: type mismatch;
[error]  found   : List[nelson.Manifest.Plan]
[error]  required: List[List[nelson.Manifest.Plan]]
[error]         c  = a.flatTraverse[List, Plan](_.toList)
[error]                                           ^
[error] one error found
```

core/src/main/scala/ManifestValidator.scala
adelbertc: Ah yeah fair, let's hold off on it for now then. Can we then modify the bits of code in the scheduler impls that currently call `.right.toOption` to instead actually inspect `Left(_)` and `Right(_)` and error out if it's a `Left(_)` so we don't silently "fail" there?

PR #153, opened 2018-10-01
Container replication for Pulsar

timperrett:


timperrett:
Once I fix something with my install I’ll test this out - it should work but I’m a little worried about the docker client compatibility

PR #154, opened 2018-10-01
Support for listing blueprints

timperrett:


core/src/main/scala/storage/h2.scala
```diff
@@ -1633,7 +1659,7 @@ final case class H2Storage(xa: Transactor[IO]) extends (StoreOp ~> IO) {
       case Blueprint.Revision.Discrete(n) => getDiscrete(n)
     }
 
-    query.map(_.map(blueprintFromRow))
+    query.map(_.map(_blueprintFromRow))
```
adelbertc: Why not just call it `blueprintFromRow` since you deleted the old one?

core/src/main/scala/storage/h2.scala
```diff
@@ -1600,15 +1601,40 @@ final case class H2Storage(xa: Transactor[IO]) extends (StoreOp ~> IO) {
...
+  private def _blueprintFromRow(row: BlueprintRow): Blueprint =
+    Blueprint(row._1, row._2, row._3, Blueprint.Revision.Discrete(row._5), Blueprint.State.Active, row._4, Template.load(s"${row._2}-${row._5}", row._6), row._7)
+
+  /**
+   * Fetches the latest blueprint revision for every known blueprint. The idea here
+   * is to provide insight into what blueprints exist in the system; given we have the
+   * inspection API for checking specific revisions, simply showing what's available and
+   * a pointer to the latest revision, should be informative enough.
+   */
+  def listBlueprints: ConnectionIO[List[Blueprint]] = {
```
adelbertc: `val`?

PR #157, opened 2018-10-06
Add support for Github Deployments 

### LINKED ISSUE #155: RFC: API for direct deployment
It has long been a design choice of Nelson to rely on Github webhook callbacks as a means to trigger deployment via Nelson. This was historically done for a few reasons:

- Having a tagged release enabled a fixed, historical point in the repository history. Tags immutably reference Git SHAs.

- The [Github Release](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) API enables the addition of "release assets". This allows Nelson to reference a checked-in, version-controlled definition of what logically should be deployed, whilst allowing CI systems to indicate the exact version they produced without having some central mutable state. 

- Leveraging releases in this way enable users to send webhooks to other systems (e.g. continuous testing), thus avoiding the need for Nelson itself to implement a sophisticated callback system.

Generally speaking, this has served the project really well, and suits poly-repo deployment environments perfectly. Even when a given repo has a few deployable systems, deploying them all together generally makes sense (for example, the same service, logically deployed internally vs at the edge, thus having different security needs). However, we do see some use cases appearing where by users would like to leverage Nelson when they have many unrelated deployable artifacts sourced from a single repo (the large / mono repo case). Here, things start to break down as the release process will - from release to release - be inconsistent. Moreover, there are distinct problems in even knowing what one should deploy.

The author fundamentally disagrees with this approach, but cannot ignore that this is a situation many organizations find themselves needing to operate with. With this frame, there is a solution we could implement that would give some of the same guarantees but rely on what i'm calling "direct deployments".

## Expanding Github Hooks

It would turn out that Github actually [has a deployment API](https://developer.github.com/v3/repos/deployments/) which is not exposed on the UI. This is primarily for triggering deployment workflows, which seems to be a decent fit for Nelson. 

**This does not replace using releases as a trigger for deployments**

The author proposes that this support of the deployments API be in addition to the existing support, allowing users to choose the path that makes most semantic sense for them. When Nelson submits the deployment request, the integration can provide a `payload` - [outlined here](https://developer.github.com/v3/repos/deployments/#simple-example) in the docs. The author supposes that this be a JSON encoding of a deployable YAML.

timperrett:
Work in progress, but the parsing support is in.

resolves #155 

- [x] Support inbound webhook with the deployment payload
- [x] Remove the dependency on downloading deployables as release assets
- [x] Figure out how we are encoding the payloads into the `payload` field. Potential options are:
  * Protobuf encode deployables, base64 binary inline
  * YAML encode deployables, base64 binary inline array (from JSON encoding perspective)
- [x] update createHook function to include pull request and deployment events


api/src/main/protobuf/nelson/version.proto
```diff
@@ -0,0 +1,27 @@
...
+//:
+//: ----------------------------------------------------------------------------
+syntax = "proto3";
+
+package nelson.api;
+
+option go_package = "nelson";
+
+message Version {
+  int32 series = 1;
```
adelbertc: I always wondered why we don't call this major, minor, and patch, per semantic versioning https://semver.org/

core/src/main/scala/Github.scala
```diff
@@ -83,35 +85,67 @@ object Github {
...
+  ) extends Event {
+    def findDeployable(withName: String): IO[Manifest.Deployable] =
+      deployables.find(_.name.trim.toLowerCase == withName.trim.toLowerCase)
+        .fold[IO[Manifest.Deployable]](IO.raiseError(new ProblematicDeployable(s"could not find an asset named $withName", url)))(IO.pure)
+  }
+
+  // NOTE(timperrett): this is legacy, but we're keeping it around so that
+  // our webhooks still give valid responses to Github in the event that
+  // operators of Nelson forget to disable the release webhook.
   final case class Release(
```
adelbertc: Mark as `@deprecated("Release triggers are deprecated in favor of Deployments.", "0.12.0")`

core/src/main/scala/Json.scala
```diff
@@ -245,8 +230,90 @@ object Json {
...
+   * }
+  */
+  implicit val GithubDeploymentEventDecoder: DecodeJson[Github.Deployment] =
+    DecodeJson(z => for {
+      a <- (z --\ "deployment" --\ "id").as[Long]
+      x <- (z --\ "repository" --\ "full_name").as[String]
+      b <- Slug.fromString(x).map(DecodeResult.ok
+           ).valueOr(e => DecodeResult.fail(e.getMessage,z.history))
+      c <- (z --\ "deployment" --\ "ref").as[String]
+      d <- (z --\ "deployment" --\ "environment").as[String]
```
adelbertc: Perhaps pull the cursor down into `deployment` into a variable so we're not constantly traversing the same key? Could do like:

```scala
for {
  deployment <- DecodeResult.ok(z --\ "deployment")
  a <- (deployment --\ "id").as[Long]
  ...
}
```

core/src/main/scala/Json.scala
```diff
@@ -245,8 +230,90 @@ object Json {
...
+      e <- (z --\ "deployment" --\ "payload").as[String]
+      f <- (z --\ "repository" --\ "id").as[Long]
+      g <- (z --\ "deployment" --\ "url").as[String]
+    } yield {
+      // NOTE(timperrett): this seems a little sketchy as we're invoking the
+      // protobuf decoder right here in the JSON decoder, even thought we've
+      // no idea if things might work here or not... we can do better.
```
adelbertc: Curious what the alternative would be? Seems like it is reasonable enough.

core/src/main/scala/Json.scala
```diff
@@ -245,8 +230,90 @@ object Json {
...
+        ref = c,
+        environment = d,
+        repositoryId = f,
+        deployables = converted,
+        url = g
+      )
+    })
+
+  // TODO(timperrett): what do we do here about encoding the assets that
+  // are shipped to us as proto format?
```
adelbertc: If proto is the only format we expect to be sent then I'd say encode it like we received it.

core/src/main/scala/Manifest.scala
```diff
@@ -405,26 +405,17 @@ object Manifest {
       .map(l => Foldable[List].fold(l))
   }
 
-  private def addVersionToLoadbalancers(m: Manifest)(r: Github.Release): List[Loadbalancer] = {
-    val major = Version.fromString(r.tagName).map(_.toMajorVersion)
+  private def addVersionToLoadbalancers(m: Manifest)(r: Github.Deployment): List[Loadbalancer] = {
+    // NOTE(timperret): given `ref` might not actually be parsable as a `Version`, something
+    // needs to happen here that is a little more meaningful.
```
adelbertc: It definitely won't always be the case right, especially if we're releasing as part of a PR/branch or for experimentation. We should definitely have some handling here for that.

core/src/main/scala/storage/op.scala
```diff
@@ -59,8 +59,8 @@ object StoreOp {
   def addUnit(unit: UnitDef @@ Versioned, repoId: ID): StoreOpF[Unit] =
     Free.liftF(AddUnit(unit, repoId))
 
-  def createRelease(repositoryId: Long, r: Github.Release): StoreOpF[Unit] =
-    Free.liftF(CreateRelease(repositoryId, r))
+  def createRelease(r: Github.Deployment): StoreOpF[Unit] =
```
adelbertc: Also rename to `createDeployment`

timperrett: We're using that elsewhere sadly

api/src/main/protobuf/nelson/version.proto
timperrett: ha, we actually did this for a reason. This way its more obvious what the fields mean - we would continually run into people who were like "my understanding of `minor` is X" so we were like "this is the feature version"

core/src/main/scala/Github.scala
timperrett: Actually we cannot do this, as we compile lifting warnings into errors.

core/src/main/scala/Json.scala
timperrett: No decent alternative came to mind, which is why im doing it here. The upside is that we get "pure" domain objects throughout the rest of the code, and there's no leak of wire code into the rest of the program

PR #173, opened 2018-10-19
Differentiate between Deployments and DeploymentEvent

### LINKED ISSUE #172: Calling back to Github results in a different deployment definition
This causes problems deserializing the payload when calling back for commit operations.

timperrett:
It would turn out - rather annoyingly - that Github has a slight protocol difference between the information you get on the webhook, vs the information you get when requesting an individual deployment directly using the API. 

This PR resolves that issue by breaking the `Github.Deployment` decoder into its own discrete decoder, and then wrapping the inbound webhook body with its own type, as the only time our code path depends upon the `repositoryId` is when we create the "release" (dang that table really needs to be renamed)

resolves #172


timperrett:
Tested in production - this is working as expected

PR #176, opened 2018-11-06
[docs] First-use guide

timperrett:


docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+    weight: 2
+---
+
+## Assumptions
+
+Whilst Nelson is typically non-perscriptive, for the unfamiliar user and for the remainder of this article, the following assumptions will be made to avoid confussion:
+
+* You can launch any container you like using Nelson; there is no requirement from the system perspective that the container behaves in a particular way other than having a [typical Nelson entrypoint](https://docs.docker.com/engine/reference/builder/#entrypoint).
+
+* Nelson is setup and running in your environment, and configured to talk to a scheudler of your choosing.
```
adelbertc: "scheduler" typo

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+  main:
+    parent: gs
+    identifier: gs-first-use
+    url: /getting-started/first-use.html
+    weight: 2
+---
+
+## Assumptions
+
+Whilst Nelson is typically non-perscriptive, for the unfamiliar user and for the remainder of this article, the following assumptions will be made to avoid confussion:
```
adelbertc: typo: non-prescriptive and confusion

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+
+* You can launch any container you like using Nelson; there is no requirement from the system perspective that the container behaves in a particular way other than having a [typical Nelson entrypoint](https://docs.docker.com/engine/reference/builder/#entrypoint).
```
adelbertc: "typical Nelson entrypoint" may be confusing, we should just say that we expect a Docker entrypoint or command that does the right thing upon being run. May also be worth noting Nelson does not allow service owners to specify additional arguments or otherwise override the command during deployment, and therefore the command should transparently work from environment to environment. This typically means using one of the Nelson environment variables like `${NELSON_ENV}` to say, render config/secrets.

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+
+* Nelson is setup and running in your environment, and configured to talk to a scheudler of your choosing.
+
+## Manifest
+
+One of the core tenets of the Nelson philosophy is that all changes are checked into source code - nothing should be actioned out of band, resutling in an untrackable system state. Nelson requires users define a manifest file and check it into the root of the source repository. This file is called `.nelson.yml` (note the preceding `.` since it is a UNIX dotfile).
```
adelbertc: typo: resulting
grammar: "Nelson requires users to define..."

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+
+This file contains the application `unit` definitions, along with any additional configuration for monitoring, alerting and scaling those units as needed. Consider this example that launches a Hello World service:
```
adelbertc: In our internal guide we have a table of some basic Nelson terminology, namely what a "unit" is versus a "stack", might be worth adding before we talk about "units"

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+Now that you have a manifest file, we can ask Nelson to enable deployments for a given repository. In practice Nelson is listening for events from Github (commits, pull requests, deployments etc) using a [repository webhook](https://developer.github.com/v3/repos/hooks/) - Nelson keeps track of all the repositories you have access to as a user on Github, but it will periodically require syncing to ensure that Nelson knows about any new repositories. The first time you use Nelson, no repositories will be listed in the system so we'll have to ask Nelson to sync them.
+
+<div class="alert alert-warning" role="alert">
+  Be aware that Nelson can only access repositories on the command line for which the provided Github access token used to <a href="/getting-started/install.html">login to Nelson</a> with, has access too.
+</div>
+
+To sync your repositories for the first time, run the following:
+
+```
+$ nelson repos sync
```
adelbertc: We need to tell users to `nelson login` first, which would also involve talking about what envvars we want.

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+Depending upon the number of organizations and repositories you have access too, the time this operation takes to complete may very. For larger organizations with thousands of repositories this may take a several minutes, but in the common case it should be relatively quick. Once completed, you can list the repositories available in your system like so:
```
adelbertc: typo: vary

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+
+```
+$ nelson repos list -o example
```
adelbertc: I usually put the longform in docs so it's clearer, e.g. `--owner`

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+--organization getnelson \
+--repository hello-world
+```
+
+This command will instruct the Nelson server to install a Github webhook on the repository specified, which will send subsequent Github Deployment events to Nelson. From here, we can 
+
+## Triggering a Deployment
+
+In most organizations where Nelson is being deployed, [continuous deployment](https://en.wikipedia.org/wiki/Continuous_delivery#Relationship_to_continuous_deployment) is an explicit goal. That is to say, merging changes into the mainline branch should trigger a deployment into the targeted datacenter(s). Nelson is integrated with Github for this purpose; listening to [Github Deployments](https://developer.github.com/v3/guides/delivering-deployments/) as a trigger for action. Whilst you are free to trigger this however you like (Nelson doesn't care, provided the invokation sends the desired [deployable specification](/getting-started/deployables.html)), but to make this easy the Nelson project provides a supporting tool called [Slipway](https://github.com/getnelson/slipway). Checkout <a href="/downloads.html">the downloads page</a> to find the latest release of Slipway and Nelson CLI.
```
adelbertc: Grammar: the sentence "Whilst you are free ..., but to make this easy ..." doesn't parse

docs/src/hugo/content/getting-started/first-use.md
```diff
@@ -0,0 +1,162 @@
...
+λ nelson stack logs 69e6ebe3efc6
+===>> logs for stack 69e6ebe3efc6
+2018-11-02T21:43:51.461Z: Pulsar workflow about to start
+2018-11-02T21:43:51.484Z: Writing policy to vault: namespace=dev unit=hello-world policy=nelson__dev__hello-world--1-0-4--416nm3q6 datacenter=us-west-2
+2018-11-02T21:43:51.529Z: Writing Kubernetes auth role 'hello-world--1-0-4--416nm3q6' to Vault...
+2018-11-02T21:43:51.636Z: Instructing us-west-2's scheduler to handle service container
+2018-11-02T21:43:52.715Z: =====> Pulsar workflow completed <=====
+```
+
+In the event there was a problem launching to the scheduler, writing to vault, or anything else encompassed by the deployment workflow: it will show here in the logs.
```
adelbertc: typo: encompassed
grammar: colon looks weird there

PR #182, opened 2018-12-08
support EC2/ECS discovery for credentials

### LINKED ISSUE #180: Allow the AWS provider chain to be configured
Currently static basic credentials are being used. In scenarios, such as running on ECS, it would be preferably to use IAM.

goedelsoup:
Fixes #180 

Currently, the AWS infrastructure configuration demands the use of basic credentials. While these can be securely provided through a number of solutions, falling back on IAM discovery can eliminate their need entirely when users are operating on a platform which supports use of the EC2 metadata service.


timperrett:
Great stuff @goedelsoup!

PR #184, opened 2018-12-09
add support for magnetar blueprints

goedelsoup:
Adds basic blueprint support for the magnetar workflow. I have tested it is generating valid nomad job specs but haven't run it against a live cluster.

TODO
- [ ] need to replace unit testing space related to blueprint injection


adelbertc:
One thing you'll definitely want to make sure of is that your template renders correctly for all cases - Scalate is............ very fiddly. Took a lot of defensive templating/hacks to make things work out well for the Canopus blueprints.

core/src/main/scala/Config.scala
```diff
@@ -532,16 +537,28 @@ object Config {
...
+
+      val creds =
+        (kfg.lookup[String]("docker.user"),
+         kfg.lookup[String]("docker.password")
+        ).mapN((a,b) => Docker.Credentials(a,b))
+
+      /*
+      If a datacenter specific registry isn't specified fall back to Nelson's publish registry
+       */
+      val registry = kfg.lookup[String]("docker.registry") getOrElse kfg.require[String]("docker-registry")
```
rodriguezsergio: You need
```
           nomad {
             endpoint = "{{ .nomad.endpoint }}"
             timeout = {{ .nomad.timeout }}
             mhz-per-cpu = {{ .nomad.mhz_per_cpu }}
              docker {
                  registry = "{{ .docker_registry }}"
              }
              docker-registry = "{{ .docker_registry }}"
           }
```
in `core/src/main/resources/nelson/defaults.cfg`.

goedelsoup: @timperrett @adelbertc Thoughts? It's been a minute now and I don't really know what I was thinking here. This was `docker.host`, which is semantically inaccurate because it is Nomad's primary Docker registry rather than a Docker execution target. `docker.registry` makes more sense between the two to me, but shouldn't Nomad just be using `[datacenter].docker-registry`?

core/src/test/resources/nelson/manifest.v1.everything.yml
```diff
@@ -91,6 +94,32 @@ plans:
     memory: 512
     instances:
       desired: 1
+    resources:
+    - ref: s3
+      uri: http://s3.aws.com
+      assume: sts:marketing
+    - ref: postgres
+      uri: postgresql://postgres.consul:5432/public
+      role: >-
```
goedelsoup: reminder to @goedelsoup to remove these demo values

docker/hashistack/docker-compose.yml
```diff
@@ -0,0 +1,125 @@
+version: '3'
```
goedelsoup: reminder to @goedelsoup this `docker/hashistack` stuff isn't stable, remove

goedelsoup:
closing this to cleanup with master and some changes in reaction to testing

PR #198, opened 2018-12-22
add documentation for development tools

goedelsoup:


docs/src/hugo/content/documentation/cli.md
```diff
@@ -355,6 +356,49 @@ There are occasions where you might want to try re-deploying a stack: for exampl
...
+```
+
+Any validation errors encountered will be reported back to you exactly as they would during the standard runtime operation of Nelson. If you would like to test a manifest whose name is different than the default `.nelson.yml`, you can supply a file path explicitly with the `-m` command line option. 
+
+### Linting Templates
+
+Debugging file templates is a notoriously difficult and frustrating component to readying the deployment of an application, regardless of the runtime and tooling which is used. Nelson aids in this process by allowing you to simulate the template rendering process exactly as it would occur during deployment.
+
+```
+λ nelson lint manifest --unit foo --resource s3 --resouce mysql --template foo.ctmpl
```
adelbertc: typo in the second 'resource'

docs/src/hugo/content/documentation/cli.md
```diff
@@ -355,6 +356,49 @@ There are occasions where you might want to try re-deploying a stack: for exampl
...
+2017/02/06 20:31:57.832769 [INFO] consul-template v0.18.0 (5211c66)
+2017/02/06 20:31:57.832781 [INFO] (runner) creating new runner (dry: true, once: true)
+2017/02/06 20:31:57.832968 [INFO] (runner) creating watcher
+2017/02/06 20:31:57.837192 [INFO] (runner) starting
+2017/02/06 20:31:57.837213 [INFO] (runner) initiating run
+Consul Template returned errors:
+/consul-template/templates/nelson8021968182946245276.template: parse: template: :3: unterminated quoted string
+```
+
+At the time of writing, Nelson only support rendering and linting Consul templates; however, since the underlying template engine executed via Docker, linting could be extended to support any engine for which a Docker container and command set can be provided.
```
adelbertc: ... since the underlying template engine **is** executed via Docker ...

PR #209, opened 2019-01-14
[breaking] refactor how configuration is loaded for schedulers

timperrett:
This PR includes a significant refactor of how we configure the infrastructure block of the config file. Specifically, the following 0.12 config:

```
infrastructure {
  scheduler {
    scheduler = "nomad"
    kubernetes {
      in-cluster = false
      timeout    = 10 seconds
      kubeconfig = "/opt/application/conf/kubeconfig"
    }
   # and nomad if configured
  }
}
``` 
This now becomes:

```
infrastructure {
  # values can be "kubernetes" | "nomad"
  scheduler = "nomad"
  # values can be "kubernetes" | "consul" | "consul:lighthouse" | "noop"
  routing = "consul:lighthouse"
  kubernetes {
    in-cluster = false
    timeout    = 3 seconds
    kubeconfig = "/opt/application/conf/kubeconfig"
  }
  # and nomad etc...
}
```

The specific change here is that the configuration is now more explicit, as we found that Nelson cannot possibly infer all combinations of the sparse matrix therein.

### DIFF
--- a/core/src/main/resources/nelson/defaults.cfg
+++ b/core/src/main/resources/nelson/defaults.cfg
@@ -238,19 +238,19 @@ nelson {
   #   texas {
   #     domain = "your.company.com"
   #     docker-registry = "registry.service.texas.your.company.com/whatever"
-
+  #
   #     infrastructure {
   #       # values can be "kubernetes" | "nomad"
   #       scheduler = "nomad"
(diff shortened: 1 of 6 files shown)


stew:
LGTM

PR #214, opened 2019-02-13
Raise hell if the names are empty or the template is bad

timperrett:
Got into a sticky situation where a user uploaded a template that was not parsable, breaking all the templating listing and inspection capabilities. This PR adds more validation to the server side to prevent these kinds of accidents

### DIFF
--- a/core/src/main/scala/Nelson.scala
+++ b/core/src/main/scala/Nelson.scala
@@ -359,13 +359,18 @@ object Nelson {
       StoreOp.findBlueprint(name, revision).foldMap(cfg.storage)
     }
 
+  /*
+   * NOTE: to prevent totally invalid templates being specified (i.e. things that make the
+   * template parser choke) we defensivly try to proof the inbound blueprint which should
+   * raise an error in the event the blueprint is junk.
+   */
   def createBlueprint(name: String, description: Option[String], sha: Sha256, template: String): NelsonK[Option[Blueprint]] =
-    Kleisli { cfg =>
+    proofBlueprint(template).flatMap(_ => Kleisli { cfg =>
       (for {
         a <- StoreOp.insertBlueprint(name, description, sha, template)
         b <- StoreOp.findBlueprint(name, Blueprint.Revision.HEAD)
       } yield b).foldMap(cfg.storage)
-    }
+    })
 
   def proofBlueprint(templateContent: String): NelsonK[String] = {
     import blueprint.{Template, EnvValue, Render}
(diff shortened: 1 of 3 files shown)


http/src/main/scala/plans/Blueprints.scala
```diff
@@ -147,7 +147,10 @@ final case class Blueprints(config: NelsonConfig) extends Default {
     case req @ POST -> Root / "v1" / "blueprints" & IsAuthenticated(session) if IsAuthorized(session) =>
       decode[BlueprintRequestJson](req){ bpr =>
         if (hasIntegrity(bpr.sha256, bpr.template))
-          json(Nelson.createBlueprint(bpr.name, bpr.description, bpr.sha256, bpr.template))
+          if (Option(bpr.name).exists(_ != ""))
+            json(Nelson.createBlueprint(bpr.name, bpr.description, bpr.sha256, bpr.template))
+          else
+            BadRequest(s"Blueprints must have a name, otherwise uesrs will not be able to reference them.")
```
drewgonzales360: Definitely done this before :sweat_smile:

drewgonzales360: s/uesrs/users/g

core/src/main/scala/Nelson.scala
```diff
@@ -359,13 +359,18 @@ object Nelson {
...
     }
 
+  /*
+   * NOTE: to prevent totally invalid templates being specified (i.e. things that make the
+   * template parser choke) we defensivly try to proof the inbound blueprint which should
+   * raise an error in the event the blueprint is junk.
+   */
   def createBlueprint(name: String, description: Option[String], sha: Sha256, template: String): NelsonK[Option[Blueprint]] =
-    Kleisli { cfg =>
+    proofBlueprint(template).flatMap(_ => Kleisli { cfg =>
```
stew: can't this just go into the for comprehension?

timperrett: The for-comp on the right is a `StoreOp`, not a `NelsonK`. At present, the `StoreOp` is lifted into the `NelsonK`, but you wont be able to compose them directly, unless im missing something or there's a different syntax to convert / lift between the two types.

stew: right sorry, I was misreading and I don't remember how to scala, lol

http/src/main/scala/plans/Blueprints.scala
```diff
@@ -147,7 +147,10 @@ final case class Blueprints(config: NelsonConfig) extends Default {
     case req @ POST -> Root / "v1" / "blueprints" & IsAuthenticated(session) if IsAuthorized(session) =>
       decode[BlueprintRequestJson](req){ bpr =>
         if (hasIntegrity(bpr.sha256, bpr.template))
-          json(Nelson.createBlueprint(bpr.name, bpr.description, bpr.sha256, bpr.template))
+          if (Option(bpr.name).exists(_ != ""))
```
lu4nm3: I would suggest doing something like this instead:

`Option(bpr.name).exists(_.trim.nonEmpty)`

This way you can account for more than the empty string (e.g. newlines, tabs, etc).

lu4nm3:
🐑

http/src/main/scala/plans/Blueprints.scala
timperrett: This is a good suggestion, thanks!

PR #216, opened 2019-02-26
Have the ability to specify an ELB as internal

timperrett:
This adds the ability to specify that the ELB that nelson launches is `internal`. For more information see: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-internal-load-balancers.html


timperrett:
PTAL @adelbertc @goedelsoup

PR #219, opened 2019-03-12
Fix bug where the 'only' filter of DC targets was not being parsed

adelbertc:


adelbertc:
Don't merge this yet.. adding tests

adelbertc:
I'm doing this on `0.11.x` first since that's what I'm on but I will follow-up with a forward-port once this goes through.

PR #222, opened 2019-03-15
fix typos

preetapan:
Fixed a few typos I noticed


timperrett:
Thank you @preetapan!

PR #228, opened 2019-03-28
k8s: Use contexts instead of namespaces for flexibility

adelbertc:
NOTE: This is a breaking change for in-cluster deployments, more info below:

Using the --namespace flag assumes the "current" credentials are valid
for that namespace (e.g. kubectl does not try to switch contexts when
you explicitly specify a namespace, which makes sense). However it is
possible that a Kubernetes deployment expects different credentials
per-namespace, which KUBECONFIG supports. However given how we're using
--namespace right now, Nelson isn't leveraging that flexibility.

This change instead uses --context to explicitly specify the context
(and therefore token + namespace). However since contexts can be named
anything (there is a logical name for each context which ties together
(cluster, namespace, token)), and because we expect each DC to have
its own KUBECONFIG, Nelson will assume the context name is the same as
the namespace name.

In addition this change also removes in/out-cluster distinction in the Kubernetes backend.

Previous in-cluster behavior used assumed administrative credentials
automatically mounted in the Pod
(https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod)
to do deployments in-cluster. However with the previous change to use
--context instead of --namespace, this no longer works (because there is
no KUBECONFIG file, it just uses the token). Therefore even if Nelson
is deployed in the same cluster a corresponding kubeconfig must still be
mounted + specified. In any case this also makes the semantics perhaps
slightly less confusing and/or more consistent.

### DIFF
--- a/core/src/main/scala/Config.scala
+++ b/core/src/main/scala/Config.scala
@@ -17,7 +17,6 @@
 package nelson
 
 import nelson.BannedClientsConfig.HttpUserAgent
-import nelson.Infrastructure.KubernetesMode
 import nelson.audit.{Auditor,AuditEvent}
 import nelson.cleanup.ExpirationPolicy
 import nelson.docker.Docker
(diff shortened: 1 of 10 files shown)


core/src/main/scala/Kubectl.scala
```diff
@@ -64,14 +64,14 @@ final class Kubectl(mode: KubernetesMode) {
...
-    exec(List("kubectl", "get", "deployment", stackName.toString, "-n", namespace.root.asString, "-o", "json"), emptyStdin)
+    exec(List("kubectl", "get", "deployment", stackName.toString, "--context", namespace.root.asString, "-o", "json"), emptyStdin)
       .flatMap(_.output)
       .flatMap { stdout =>
         IO.fromEither(Parse.decodeEither[DeploymentStatus](stdout.mkString("\n")).leftMap(kubectlJsonError))
       }
 
   def getCronJob(namespace: NamespaceName, stackName: StackName): IO[JobStatus] =
-    exec(List("kubectl", "get", "job", "-l", s"stackName=${stackName.toString}", "-n", namespace.root.asString, "-o", "json"), emptyStdin)
+    exec(List("kubectl", "get", "job", "-l", s"stackName=${stackName.toString}", "--context", namespace.root.asString, "-o", "json"), emptyStdin)
```
timperrett: actually this is *very* breaking... this assumes that the kubeconfig files people are using are structured in a particular manner, correct? This would break my setup today.

core/src/main/scala/Config.scala
```diff
@@ -542,14 +541,10 @@ object Config {
         })
     }
 
-    def readKubernetesOutClusterParams(kfg: KConfig): Option[KubernetesMode] =
-      kfg.lookup[String]("kubeconfig").map(kubeconfig => KubernetesMode.OutCluster(Paths.get(kubeconfig)))
-
     def readKubernetesInfrastructure(kfg: KConfig): Option[Infrastructure.Kubernetes] = for {
-      inCluster <- kfg.lookup[Boolean]("in-cluster")
-      mode      <- if (inCluster) Some(KubernetesMode.InCluster) else readKubernetesOutClusterParams(kfg)
```
timperrett: I think this will also break any existing configuration files in the field.

timperrett:
@adelbertc did you get a chance to think about what to do here? This will break existing deployments

PR #233, opened 2019-04-29
Create PKI roles so services can create certificates

drewgonzales360:
In Vault, the [PKI engine](https://www.vaultproject.io/api/secret/pki/index.html) requires that a role be [created](https://www.vaultproject.io/api/secret/pki/index.html#create-update-role) so that certificates can be issued against it. When certs are issued, services that take ingress traffic can TLS secured. 

This PR adds functionality to create these roles.

### DIFF
--- a/core/src/main/scala/vault/http/Http4sVault.scala
+++ b/core/src/main/scala/vault/http/Http4sVault.scala
@@ -61,6 +61,8 @@ final class Http4sVaultClient(
     case ct: CreateToken          => createToken(ct)
     case kr: CreateKubernetesRole => createKubernetesRole(kr)
     case dr: DeleteKubernetesRole => deleteKubernetesRole(dr)
+    case cpkir: CreatePKIRole     => createPKIRole(cpkir)
+    case dpkir: DeletePKIRole     => deletePKIRole(dpkir)
   }
 
   val log = Logger[this.type]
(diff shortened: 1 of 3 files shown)


core/src/main/scala/vault/op.scala
```diff
@@ -130,6 +130,15 @@ object Vault {
     roleName: String
   ): VaultF[Unit] = Free.liftF(DeleteKubernetesRole(authClusterName, roleName))
 
+  // TODO(drewgonzales360): Do some sort of validation to ensure the roleName is a unitName
```
timperrett: Remove this line, as there's no validation here - the vault subsystem is a generic vault client.

core/src/main/scala/vault/http/Http4sVault.scala
```diff
@@ -61,6 +61,8 @@ final class Http4sVaultClient(
     case ct: CreateToken          => createToken(ct)
     case kr: CreateKubernetesRole => createKubernetesRole(kr)
     case dr: DeleteKubernetesRole => deleteKubernetesRole(dr)
+    case cpkir: CreatePKIRole     => createPKIRole(cpkir)
```
timperrett: Where is this type defined?

core/src/main/scala/vault/http/json.scala
```diff
@@ -114,4 +114,13 @@ trait Json {
       ("policies" :?= kr.policies) ->?:
       jEmptyObject
     }
+
+    implicit val jsonCreatePKIRole: EncodeJson[CreatePKIRole] =
+    EncodeJson { cpkir =>
```
timperrett: Indent this block

PR #244, opened 2019-08-23
Fix PKI path for role creation

miranp-lyft:
PKI path from change #242 incorrectly specifies the Vault path for PKI role creation; this change fixes that path, and assumes a default path of `pki` otherwise (Vault default). 

If the user is configuring a custom path, it should be specified in the Nelson configuration under:
```
datacenters {
    <datacenter> {
       ...

      policy {
        ...

        # Path to your PKI backend. Optional. If specified, unit will get
        # create and update capabilities on ${pki-path}/issue
        pki-path = "pki/my-path/dev"
      }
    }
  }
```

and _not_ under the `infrastructure { vault { ... } }` subconfiguration.

### DIFF
--- a/core/src/main/scala/Workflow.scala
+++ b/core/src/main/scala/Workflow.scala
@@ -162,6 +162,7 @@ object Workflow {
 
     def writePKIRoleToVault(dc: Datacenter, sn: StackName): WorkflowF[Unit] =
       Vault.createPKIRole(
+        pkiPath = dc.policy.pkiPath,
         engineName = dc.name,
         roleName = sn.toString,
         serviceAccountNames = List(sn.toString),
@@ -172,6 +173,7 @@ object Workflow {
 
     def deletePKIRoleFromVault(dc: Datacenter, sn: StackName): WorkflowF[Unit] =
       Vault.deletePKIRole(
+        pkiPath = dc.policy.pkiPath,
         engineName = dc.name,
         roleName = sn.toString
       ).inject
--- a/core/src/main/scala/vault/http/Http4sVault.scala
+++ b/core/src/main/scala/vault/http/Http4sVault.scala
@@ -154,12 +154,12 @@ final class Http4sVaultClient(
     authBackendPrefix.map(_ + cn).getOrElse(cn)
 
   def createPKIRole(cpkir: CreatePKIRole): IO[Unit] = {
-    val engine = kubernetesAuthEngineName(cpkir.engineName)
-    reqVoid(IO.pure(Request(POST, v1BaseUri / engine / "roles" / cpkir.roleName)))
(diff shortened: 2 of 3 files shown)


core/src/main/scala/vault/http/Http4sVault.scala
```diff
@@ -154,12 +154,12 @@ final class Http4sVaultClient(
     authBackendPrefix.map(_ + cn).getOrElse(cn)
 
   def createPKIRole(cpkir: CreatePKIRole): IO[Unit] = {
-    val engine = kubernetesAuthEngineName(cpkir.engineName)
-    reqVoid(IO.pure(Request(POST, v1BaseUri / engine / "roles" / cpkir.roleName)))
+    val pkiPath = cpkir.pkiPath.getOrElse("pki")
```
miranp-lyft: happy to move this instantiation higher up in the hierarchy of config initialization if it makes more sense~

core/src/main/scala/vault/op.scala
```diff
@@ -131,22 +131,25 @@ object Vault {
   ): VaultF[Unit] = Free.liftF(DeleteKubernetesRole(authClusterName, roleName))
 
   def createPKIRole(
+    pkiPath: Option[String],
```
timperrett: As a general comment, which we should change here, i generally add new arguments to a function at the end of the argument list as there are cases through the codebase where we sometimes rely on default params with multiple params of the same type (gotta love strings, for example). Whilst this is largely historical, it avoids accidentally changing functionality that relies on default params.

core/src/main/scala/Workflow.scala
```diff
@@ -162,6 +162,7 @@ object Workflow {
 
     def writePKIRoleToVault(dc: Datacenter, sn: StackName): WorkflowF[Unit] =
       Vault.createPKIRole(
+        pkiPath = dc.policy.pkiPath,
```
timperrett: I'm not 100% sure this is what you wanted. Right here this is going to give you the literal value including the `%` template values. I think what you in fact want would be something that utilizes some of the functions here: https://github.com/getnelson/nelson/blob/master/core/src/main/scala/vault/policies.scala#L27 - you may also want to consider writing a function for this particular interpolation, as it looks like you might need a path that is ever so slightly different to the ones being used in the policy?

miranp-lyft: OH okay, that makes more sense - I was under the impression that this was something that was configured before-Nelson 😅 will take a second look~

timperrett: Yeah so it is configured prior to bootup, but because of the need to template these paths, they have to be interpolated at runtime. The config only holds the templatized values 🥾

PR #245, opened 2019-10-16
AWS DCs Use Network Load Balancers

drewgonzales360:
# Summary

I'm trying to switch the [Classic Loadbalancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) for [Network Loadbalancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html).

This involves changing the java library to `aws-java-sdk-elasticloadbalancingv2`.

I have to
1. Create the NLB
2. Create `N` target groups
3. Create `N` Listeners on the NLB. Each listener maps 1:1 with a target group
4. register the ASG with _every_ target group (I know, confusing)


drewgonzales360:
@timperrett @miranp-lyft @lu4nm3

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -75,32 +77,100 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
   def loadbalancerName(name: String, v: MajorVersion, hash: String) =
-    s"$name--${v.minVersion.toExternalString}--${hash}"
+    s"$name-${v.minVersion.toExternalString}-${hash}"
 
-  def deleteELB(name: String): IO[Unit] =
+  def deleteELB(lbarn: String): IO[Unit] =
     IO {
-      val req = new DeleteLoadBalancerRequest(name)
-      cfg.elb.deleteLoadBalancer(req)
+      cfg.elb.deleteLoadBalancer(
```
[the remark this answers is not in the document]
drewgonzales360: Sounds good :smile:  Which is pure and which is impure?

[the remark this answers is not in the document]
timperrett: Ah, yes 😄 In this case:

```
val req = new DeleteLoadBalancerRequest().withLoadBalancerArn(lbarn)
```
This is pure, as there are no effects happening here, no I/O etc.

```
cfg.elb.deleteLoadBalancer(req)
```

This is impure, as it utilizes network I/O, so here, we wrap it in `IO`. As a general rule, just for readability reasons, I think its always nice to split these, as its more readable and if you want to, say, `println` the value, then simply assigning a value that will be GC'd is totally fine.

core/src/main/scala/ManifestValidator.scala
```diff
@@ -91,7 +91,7 @@ object ManifestValidator {
     units: List[UnitDef],
     whiteList: Option[ProxyPortWhitelist]): Valid[Unit] = {
 
-    // NOTE: aws allows elb names to be at most 32 charasters. Nelson generates a name for
+    // NOTE: aws allows nlb names to be at most 32 charasters. Nelson generates a name for
```
timperrett: Is this accurate for NLB? I expect it is, but lets check

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+    } yield ()
   }
 
   def launch(lb: Manifest.Loadbalancer, v: MajorVersion, ns: NamespaceName, p: Plan, hash: String): IO[DNSName] = {
     val name = loadbalancerName(lb.name, v, hash)
-    log.debug(s"caling aws client to launch $name")
+    log.debug(s"calling aws client to launch $name")
     val size = asgSize(p)
-    createELB(name, lb.routes.map(_.port), cfg.lbScheme) <* createASG(name, ns, size)
+
```
timperrett: Remove whitespaces

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
   }
 
+  // The target group has this restriction where
+  // the name of the target group can't have "double dashes". I want the loadbalancer name
+  // to match the target group name, so the loadbalancer won't have the double dashes either.
   def loadbalancerName(name: String, v: MajorVersion, hash: String) =
-    s"$name--${v.minVersion.toExternalString}--${hash}"
+    s"$name-${v.minVersion.toExternalString}-${hash}"
+
+  def getLoadBalancerArn(name: String): IO[String] = {
```
timperrett: Perhaps:

```
def getLoadBalancerArn(name: String): IO[String] = {
  val describeLoadBalancerRequest = new DescribeLoadBalancersRequest().withNames(name)
  (for {
    a <- IO(cfg.nlb.describeLoadBalancers(describeLoadBalancerRequest))
    b  = a.getLoadBalancers.asScala.toList
    c  = loadbalancerList.head.getLoadBalancerArn
  } yield ()).recoverWith{
    case e: Exception =>
      log.error(s"aws call to get loadbalancer ARN ${describeLoadBalancerRequest.getNames.asScala.head} failed.")
      IO.raiseError(FailedLoadbalancerDeploy(s"could not find loadbalancer and therefore could not delete loadbalancer", e.getMessage))
  }
}
```

I haven't tried this, but something along those lines to avoid the map map map situation

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+        log.error(s"aws call to get loadbalancer ARN ${describeLoadBalancerRequest.getNames.asScala.head} failed.")
+        IO.raiseError(FailedLoadbalancerDeploy(s"could not find loadbalancer and therefore could not delete loadbalancer", e.getMessage))
+    }
+  }
 
-  def deleteELB(name: String): IO[Unit] =
-    IO {
-      val req = new DeleteLoadBalancerRequest(name)
-      cfg.elb.deleteLoadBalancer(req)
+  def deleteNLB(lbarn: String): IO[Unit] = {
```
timperrett:
```
def deleteNLB(lbarn: String): IO[Unit] =
  IO {
    val req: DeleteLoadBalancerRequest = new DeleteLoadBalancerRequest()
      .withLoadBalancerArn(lbarn)

    cfg.nlb.deleteLoadBalancer(req)
    ()
  }
```

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+        .withType(LoadBalancerTypeEnum.Network)
         .withScheme(scheme.toString)
-        .withSecurityGroups(cfg.elbSecurityGroupNames.toList.asJava)
         .withSubnets(cfg.availabilityZones.map(_.publicSubnet).asJava)
 
-      val resp = cfg.elb.createLoadBalancer(req)
+      val resp = cfg.nlb.createLoadBalancer(req)
+      val loadbalancers: List[LoadBalancer] = resp.getLoadBalancers.asScala.toList
+
+      loadbalancers.head
```
timperrett: This still makes me uncomfortable. @adelbertc got any suggestions here to avoid using `head`? We could do some gymnastics wiith `headOption` and monad transformers, but im not sure its worth it.

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+  }
+
+  def getTargetGroupsForNLB(lbarn: String): IO[Vector[TargetGroup]] = {
+    val req = new DescribeTargetGroupsRequest()
+      .withLoadBalancerArn(lbarn)
 
-      resp.getDNSName()
+    IO(
+      cfg.nlb.describeTargetGroups(req)
+    ).map(resp => resp.getTargetGroups.asScala.toVector)
```
timperrett: Use list in place of veector

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+
+    val ctgreqs: Vector[CreateTargetGroupRequest] = p.map(port => createRequest(port, name))
+
+    // (@timperret): I can't think of a way to import this. When I put it at the top,
+    // I got all kinds of weird errors saying parameter is "never used".
+    import cats.implicits._
+    ctgreqs.traverse(req => createSingleTargetGroup(req))
+  }
+
+  def getTargetGroupsForNLB(lbarn: String): IO[Vector[TargetGroup]] = {
```
timperrett: List not vector

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +65,179 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+    val req = new DescribeTargetGroupsRequest()
+      .withLoadBalancerArn(lbarn)
 
-      resp.getDNSName()
+    IO(
+      cfg.nlb.describeTargetGroups(req)
+    ).map(resp => resp.getTargetGroups.asScala.toVector)
+  }
+
+  def deleteTargetGroups(targetGroups: Vector[TargetGroup]): IO[Unit] = {
```
timperrett: List not vector

core/src/main/scala/ManifestValidator.scala
drewgonzales360: Yup this is still true.

core/src/main/scala/loadbalancers/Aws.scala
drewgonzales360: I've used that map/map pattern a few places. This looks way nice :smile:

core/src/main/scala/loadbalancers/Aws.scala
goedelsoup: This would do the trick, but unsure how much better it is.
```
Alternative[Option]
  .guard(loadbalancers.size === 1)
  .flatMap(loadbalancers.headOption)
  .fold(IO.raiseError(...))(_.pure[IO])
```

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -68,40 +66,181 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+  }
 
-      resp.getDNSName()
+  def createTargetGroups(targetGroupNamePrefix: String, p: List[Port], vpcID: String): IO[List[TargetGroup]] = {
+
+    def createRequest(p: Manifest.Port): CreateTargetGroupRequest =
+      new CreateTargetGroupRequest()
+        .withPort(p.port)
+        .withProtocol(ProtocolEnum.TCP)
+        .withName(targetGroupNamePrefix+"-"+ p.port.toString())
```
timperrett:
```suggestion
        .withName(s"${targetGroupNamePrefix}-${p.port.toString}")
```

PR #246, opened 2019-11-05
increase the max limit

timperrett:
Users have hit the 500 limit in the field, which is entirely arbitrary as its really for validation purposes than anything else, so we're not artificially restricting it

PTAL @adelbertc @miranp-lyft

### DIFF
--- a/core/src/main/scala/yaml/ManifestV1Parser.scala
+++ b/core/src/main/scala/yaml/ManifestV1Parser.scala
@@ -127,7 +127,7 @@ object ManifestV1Parser {
   }
 
   def validateInstances(i:Int): YamlValidation[Int] =
-    validatePlanConstraints(i, invalidInstances)(i => i > 0 && i < 500)
+    validatePlanConstraints(i, invalidInstances)(i => i > 0 && i < 5000)
 
   def validateVolumes(volume: VolumeYaml): YamlValidation[Volume] = {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/yaml/ManifestV1Parser.scala
```diff
@@ -127,7 +127,7 @@ object ManifestV1Parser {
   }
 
   def validateInstances(i:Int): YamlValidation[Int] =
-    validatePlanConstraints(i, invalidInstances)(i => i > 0 && i < 500)
+    validatePlanConstraints(i, invalidInstances)(i => i > 0 && i < 5000)
```
adelbertc: Is this just validating that the number of instances/replicas of a workload is 0 < i < 5000? Why have an upper bound at all?

timperrett:
@miranp-lyft its flakey because there was a change in time precision between JDK 8 and JDK 9, so it would seem 😢 I made a chance on master that should fix this to use JDK8 for now

core/src/main/scala/yaml/ManifestV1Parser.scala
timperrett: @adelbertc at the time we wanted a way to ensure that the number people specified was "sane", so that people didn't just ask for more and more replicas (like a million, for example). WDYT?

PR #247, opened 2019-11-06
AWS DC: Try to remove ELB if NLB doesn't exist

drewgonzales360:
# Summary

If Nelson loadbalancers were provisioned with an older version of Nelson, classic AWS loadbalancers were created. This PR makes deleting those loadbalancers possible with the newer versions of Nelson IE 0.15.x

# Tested
I hopped into the repl with a few scripts and checked out a commit of Nelson that created classic loadbalancers. I created a classic loadbalancer from Nelson.

Then I checked out the head of this branch which uses NLB's. I called loadbalancers.delete and watched Nelson fail to delete the NLB, then try to delete the ELB and succeed. 

```
01:15:12.394 [run-main-1] ERROR nelson.loadbalancers.Aws - aws call to get loadbalancer ARN classic--1-0-0--hashhash failed.
01:15:12.395 [run-main-1] INFO  nelson.loadbalancers.Aws - couldn't delete NLB because Loadbalancer deployment (could not find loadbalancer and therefore could not delete loadbalancer) failed because Load balancers '[classic--1-0-0--hashhash]' not found (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: LoadBalancerNotFound; Request ID: adsfsadfasdf-asdfasdf-asdfasdfsadf). attempting to delete ELB
```

I also tested creating and deleting NLB's

## Our Current Use Case
We used to use a version of Nelson that provisioned classic loadbalancers. To use the "Network Loadbalancers", we upgraded the java sdk to use the loadbalancingv2 package. This package can't delete "classic loadbalancers" ie it is not backwards compatible.

### DIFF
--- a/core/src/main/scala/loadbalancers/Aws.scala
+++ b/core/src/main/scala/loadbalancers/Aws.scala
@@ -79,7 +79,7 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
 
   def launch(lb: Manifest.Loadbalancer, v: MajorVersion, ns: NamespaceName, p: Plan, hash: String): IO[DNSName] = {
     val name = loadbalancerName(lb.name, v, hash)
-    val tgPrefix = targetGroupNamePrefix(lb.name, v)
+    val tgPrefix = targetGroupNamePrefix(lb.name, hash)
     val size = asgSize(p)
 
     for {
@@ -97,13 +97,14 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
   // The target groups have a restriction such that name of the target group can't
   // have "double dashes" and the entire name must be less than 32 characters.
   // Unfortunately, target groups will not have the same name as the loadbalancers.
-  // An example target group will be named stack-1-0-0-443 where the last number is the
-  // port that it listens on (instead of the hash). This allows stack-1-0-0-80 to be
+  // An example target group will be named stack-aevaif-443 where the last number is the
+  // port that it listens on (instead of the hash). This allows stack-qweri-80 to be
   // a different target group for the same loadbalancer. Since target groups are only
   // associated with a single loadbalancer at a time, it'll be easy to reason about which target
-  // groups are related to which loadbalancers.
-  def targetGroupNamePrefix(name: String, v: MajorVersion) =
-    s"$name-${v.minVersion.toExternalString}"
+  // groups are related to which loadbalancers. Administrators will need to reason about
+  // which target group belongs to which loadbalancer based on the hash in the target group name.
+  def targetGroupNamePrefix(name: String, hash: String) =
+    s"$name-$hash"
 
   def getLoadBalancerArn(name: String): IO[String] = {
     val describeLoadBalancerRequest = new DescribeLoadBalancersRequest().withNames(name)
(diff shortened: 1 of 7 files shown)


core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -66,28 +66,33 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
+      _ <- deleteNLB(lbarn)
       _ <- deleteTargetGroups(targetGroups)
 
       _ <- deleteASG(name)
-    } yield ()
+    } yield ()).recoverWith{
+      case e: Exception =>
+        log.info(s"couldn't delete NLB because ${e.getMessage}. attempting to delete ELB")
+        for {
+          _ <- deleteELB(name)
```
timperrett: What happens if these fail?

drewgonzales360: Errors will get logged, but Nelson will still mark the resources as terminated. Not much we can do :man_shrugging:

drewgonzales360:
@lu4nm3 @timperrett

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -97,35 +92,82 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
...
       r <- IO(cfg.nlb.describeLoadBalancers(describeLoadBalancerRequest))
       x = r.getLoadBalancers.asScala.toList
       a = x.head.getLoadBalancerArn
-    } yield a).recoverWith{
+    } yield Option(a)).recoverWith{
       case e: Exception =>
-        log.error(s"aws call to get loadbalancer ARN ${describeLoadBalancerRequest.getNames.asScala.head} failed.")
-        IO.raiseError(FailedLoadbalancerDeploy(s"could not find loadbalancer and therefore could not delete loadbalancer", e.getMessage))
+        log.error(s"aws call to get loadbalancer ARN ${describeLoadBalancerRequest.getNames.asScala.head} failed with ${e.getMessage}")
+        IO(None)
```
timperrett: Does it semantically make sense to have an `Option` return here? It's unclear to me what successful result scenario would be `None`.

drewgonzales360: Yeah I have no idea. If the network call is successful, then I want to return the string. If it's not successful, I'm not sure what to return.

In my mind, I was thinking that I want to return _something_ if the call is successful and _nothing_ if it's not. `Option` seemed like the way to go. What did you have in mind?

core/src/main/scala/loadbalancers/Aws.scala
```diff
@@ -97,35 +92,82 @@ final class Aws(cfg: Infrastructure.Aws) extends (LoadbalancerOp ~> IO) {
   // The target groups have a restriction such that name of the target group can't
   // have "double dashes" and the entire name must be less than 32 characters.
   // Unfortunately, target groups will not have the same name as the loadbalancers.
-  // An example target group will be named stack-1-0-0-443 where the last number is the
-  // port that it listens on (instead of the hash). This allows stack-1-0-0-80 to be
+  // An example target group will be named stack-aevaif-443 where the last number is the
```
timperrett: Why would the revision not be a part of the name?

core/src/main/scala/loadbalancers/Aws.scala
timperrett: In this case, one would not usually leverage `Option`; here we would use `Either` where the result is _either_ a `String` or `Error`, but the resulting type is clear in either branch (semantically, and practically). In essence, this is what you're getting with `IO` - your computation either succeeds or it doesn't, but the semantics of IO are specifically for IO-related things (which you have in this case).

Does that help?

drewgonzales360:
Closing this in favor of another PR. I have a private fork where this feature is working.

PR #250, opened 2019-12-27
updates to vault api

drewgonzales360:
# Summary
- bugfix: When Nelson attempted to create a vault role in the PKI engine, the json body wasn't being passed. When you make a request without a body, the Vault API will assume all the defaults. 
- modified the Vault API defaults a bit to rely on the trust chain instead of the domain. We don't force users to put a domain by default and allow any CName to get put on the certificate.


drewgonzales360:
There isn't any activity going on here. Gonna close

PR #251, opened 2019-12-27
Delete Classic Loadbalancer if it is present

drewgonzales360:
# Summary
When the changes to use the Network Loadbalancer were made, I changed that whole subsystem to assume _only_ AWS network loadbalancers were used. In reality, our deployment of Nelson previously used AWS Classic Loadbalancers.

If an org used Classic Loadbalancers in the past, then we should attempt to delete the classic loadbalancer when deleting the Network loadbalancer fails.


drewgonzales360:
Gonna close this from inactivity.

PR #253, opened 2020-04-02
Update manifest.md (minor typo)

song-william:
I think this should say `prod` instead of `qa`.


song-william:
@drewgonzales360 @timperrett I think I found a minor typo when reading the nelson docs.

timperrett:
Thanks for the PR. However, this isn't in fact a typo - the namespaces are arbitrary and depend on how Nelson was setup by the administrator within a given org. In $JOB we operated with a `qa` namespace, so it depends on the setup.

PR #257, opened 2020-06-17
make the thread pool sizing configurable

timperrett:
This adds the ability to specify `nelson.pools.default-size` and `nelson.pools.scheduling-size` in the nelson.cfg configuration file, adjusting the number of available threads to match the hardware that Nelson operates on, which is important for larger setups. 

```
    # size of the fixed executor that is used for many of the
    # async operations that happen internally with nelson.
    # in most cases this default should be more than sufficient
    # but as a general rule of thumb, you would want this value
    # to be number of cpus * 2
    default-size = 8

    # number of threads available for the scheduling thread pool.
    # this value should not be increased above 4 without understanding
    # what the impact of the change would be.
    scheduling-size = 4
```

### DIFF
--- a/core/src/main/resources/nelson/defaults.cfg
+++ b/core/src/main/resources/nelson/defaults.cfg
@@ -16,6 +16,20 @@
 ##: ----------------------------------------------------------------------------
 nelson {
 
+  pools {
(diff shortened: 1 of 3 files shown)


wzpher:
🚢

